PR #2, opened 2016-02-19
Share scheme definition for AppAuth

iainmcgin:


WilliamDenniss:
Shared scheme added to support bitrise.io continuous integration.  LGTM.

PR #3, opened 2016-02-20
Allow all PKCE parameters to be overwritten (not just the code verifier)

WilliamDenniss:
Previously we only allowed the `code_verifier` to be changed by the developer, but we had no way to change the `code_challenge` and `code_challenge_method` (aside from turning PKCE it off entirely by passing `nil` as the code verifier).

This patch enables all PKCE params to be modified, while retaining the opinionated S256 default.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -327,7 +327,9 @@ SEARCH_INCLUDES        = YES
 INCLUDE_PATH           = 
 INCLUDE_FILE_PATTERNS  = 
 PREDEFINED             = NS_UNAVAILABLE \
-                         "NS_ENUM(type, name)=enum name"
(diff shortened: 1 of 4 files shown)


WilliamDenniss:
Here's a gist for how to send a `plain` PKCE challenge: https://gist.github.com/WilliamDenniss/941d9b2bc62d5f7ffac4

Note that PKCE recommends S256 for clients capable of generating SHA256 hashes (which iOS clients are).  It is NOT RECOMMENDED to use `plain` PKCE with AppAuth for that reason.

The changes in this pull request can also be used to support new code challenge methods in the future.

pm-ping:
LTGM. Tested against PingFederate 8.0.4 (current release) which currently supports "plain" method only.

PR #22, opened 2016-05-20
Add support for dynamic client registration.

zamzterz:
(Sorry for the large changeset.)

Similar implementation to the one for Android:
- stores the registration response in the AuthState
- constructs the registration request body as JSON
- checks conditions on the registration response
- adds implementation to the example app

---

This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/22)

PR #24, opened 2016-05-24
Client authentication at the token endpoint

zamzterz:
This work is based off PR #22, so the actual diff is much smaller (I'll happily rebase this branch when (if?) that change is merged).

Similar implementation to the one for Android:
- defines an interface for the client authentication which is applied in `TokenRequest.URLRequestWithClientAuthentication`.
- encode token requests according to [spec](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1), replacing space character with plus sign.
- adds implementation to the example app.

One issue left to resolve:
- Right now the auto token exchange won't work if the token endpoint expects client authentication (with a dynamically registered confidential client). Should `AuthState.authStateByPresentingAuthorizationRequest` be "overloaded" to have a signature also accepting an `OIDClientAuthentication` instance?

---

This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/24)


rebeckag:
Any update on this?

rebeckag:
I'll rebase this against latest master once #22 has been merged to get a clean diff.

WilliamDenniss:
Closing in favor of #100 which is a different implementation.

Thanks for this work, I incorporated some of into #100 :-)

Issue #20, opened 2016-05-12
OSX support


nubbel:
Hey!

We're working on OSX support over at https://github.com/equinux/AppAuth-iOS.
Is there any chance we could get our changes merged into this repo (and change the name to AppAuth only)?
Alternatively, we could extract the platform-independent parts into a separate pod AppAuth-Core and create a new pod AppAuth-Mac that depends on it.

What do you think?

/cc @DarkDust

iainmcgin:
Hey! I think it would probably make sense to try and split platform-specific pieces out into separate repos, so I would vote for some "core" repo (AppAuth-Darwin-Core?), with this repo and AppAuth-MacOS depending on it / acting as plugins for it. @WilliamDenniss, any thoughts?

I had a quick look at your changes, looks like you're using WKWebView - we strongly discourage this for a few reasons:

1. It is not clear to the user what page they are actually interacting with - phishing via WKWebView is trivial, and it trains users to blindly trust login forms that may not be safe.
2. It allows a malicious / compromised app to steal user credentials easily even when a legitimate login page is presented, by injecting JS into the view that can capture credentials.
3. Users will not get a real "single sign-on" experience as the WKWebView has no access to the cookies of the main browser, so they will be required to enter their credentials again. This can be particularly problematic if a hardware second factor is required, as it may not be possible to interact with this from a WKWebView depending on how it works.

I would instead recommend you start a localhost web server to capture the OAuth response, and then send the openURL request to the default browser. This solves all three issues listed above. There may be less crude ways of handling the response - I don't know if there is an equivalent to custom scheme registration or universal links in Mac OS, but some cursory searching hasn't turned anything up.

nubbel:
Thanks for your quick answer!

I'd suggest AppAuth-Cocoa for the "core" project. You could also keep the plugins in the same repo and make them available as sub-pods.

The issues you mentioned are all valid, but are no big deal in our use case. Our app uses AppAuth to authenticate against our own identity provider, basically just replacing a native login dialog. True SSO is not required as we want the user to explicitly log in once. When the core functionality is factored out, it would become trivial for us to build a custom plugin (AppAuth-Mac-SuperDangerous?) that handles our special case.

In contrast, the "official" AppAuth-Mac plugin could implement the flow you described above. The response can be handled with a custom URL scheme just like on iOS. Universal links are not yet supported in OSX, but who knows what WWDC16 will bring ;).

Let me know if you want my help to make the changes. 🍻

[...]

PR #28, opened 2016-06-29
Add modal presentation options for SFSafariVC

maxsz:
Adds support for specifing the modal presentation and transition styles of the presented SFSafariViewController.

---

This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/28)

### DIFF
--- a/Source/OIDAuthState.h
+++ b/Source/OIDAuthState.h
@@ -136,6 +136,26 @@ typedef void (^OIDAuthStateAuthorizationCallback)(OIDAuthState *_Nullable authSt
     presentingViewController:(UIViewController *)presentingViewController
                     callback:(OIDAuthStateAuthorizationCallback)callback;
 
+/*! @fn authStateByPresentingAuthorizationRequest:presentingViewController:modalPresentationStyle:modalTransitionStyle:callback:
+    @brief Convenience method to create a @c OIDAuthState by presenting an authorization request
+        and performing the authorization code exchange in the case of code flow requests.
+    @param authorizationRequest The authorization request to present.
+    @param presentingViewController The view controller from which to present the
+        \SFSafariViewController.
+    @param modalPresentationStyle Modal presentation style for the \SFSafariViewController.
+    @param modalTransitionStyle Modal transition style for the \SFSafariViewController.
+    @param callback The method called when the request has completed or failed.
+    @return A @c OIDAuthorizationFlowSession instance which will terminate when it
+        receives a @c OIDAuthorizationFlowSession.cancel message, or after processing a
+        @c OIDAuthorizationFlowSession.resumeAuthorizationFlowWithURL: message.
+ */
++ (id<OIDAuthorizationFlowSession>)authStateByPresentingAuthorizationRequest:
+    (OIDAuthorizationRequest *)authorizationRequest
+    presentingViewController:(UIViewController *)presentingViewController
+      modalPresentationStyle:(UIModalPresentationStyle)modalPresentationStyle
+        modalTransitionStyle:(UIModalTransitionStyle)modalTransitionStyle
+                    callback:(OIDAuthStateAuthorizationCallback)callback;
+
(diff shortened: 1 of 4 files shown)


WilliamDenniss:
I believe this is a feature we want. May need to rebase on pull request #30 once that's done.

Should we add [preferredBarTintColor](https://developer.apple.com/reference/safariservices/sfsafariviewcontrollerconfiguration/1648469-preferredbartintcolor?language=objc) and [tintColor](https://developer.apple.com/reference/uikit/uiview/1622467-tintcolor?language=objc) options too?

Can you [sign the CLA](https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md)?

maxsz:
Great 👍. 

As there might be more options and other things we might want to configure on the view controller, I think there are two possible solutions. 

1. Create a "style" class/struct that has all the properties and pass it in (this is better than having more and more parameters to the already very long methods like `authStateByPresentingAuthorizationRequest :presentingViewController:modalPresentationStyle:modalTransitionStyle:callback:`)

2. Pass a closure to the methods which returns the configured SFSafariViewController. This is the most flexible option, as we won't need to adapt/change the API for new iOS versions when new properties become available (like the `preferredBarTintColor ` of `SFSafariViewControllerConfiguration` you mentioned, which will be added in iOS 10). I'm thinking something like this:
```objective-c
 - (void)presentSafariViewControllerWithViewController:(UIViewController *)parentViewController
                                       configuration:(SFSafariViewController (^)(void))configurationBlock
                                              callback:(OIDAuthorizationCallback)authorizationFlowCallback {
    _pendingauthorizationFlowCallback = authorizationFlowCallback;
    NSURL *URL = [_request authorizationRequestURL];
    if ([SFSafariViewController class]) {
      SFSafariViewController *safariVC;
      if (configurationBlock) {
        safariVC = configurationBlock();
      }
      if (!safariVC) {
       safariVC = [[SFSafariViewController alloc] initWithURL:URL
                                                       entersReaderIfAvailable:NO];
      }
      safariVC.delegate = self;
      _safariVC = safariVC;
      [parentViewController presentViewController:safariVC animated:YES completion:nil];
    }
   ...
}
```

What would you guys prefer? Do you have other options in mind?

Regarding the CLA: I'm acting on behalf of equinux AG and we've already signed the CLA (https://github.com/openid/AppAuth-iOS/issues/20#issuecomment-229709325).

StevenEWright:
I have proposed an alternative here: [https://github.com/openid/AppAuth-iOS/pull/34](https://github.com/openid/AppAuth-iOS/pull/34)

WilliamDenniss:
Firstly, thank you @maxsz for starting this discussion, it's definitely a feature that's lacking in AppAuth and I appreciate your ideas and code! 

So we have three alternatives suggested so far:
1. Add a param in the constructor for everything (this PR, #28).
2. Add a configuration block.
3. Add a factory (#34).

I'm inclined to agree with your assessment regarding point 1), that we'll end up with really long methods. At a minimum we'd want tintColor, and the list could grow quite long, and it seems like the less favored option.

Regarding 2) I think that's a nice solution for one-offs (less code than creating a class per 3)), but I also think it feels less reusable. Let's say AppAuth offers 2 styles, Apple's default (blue tint, etc), and the AppAuth opinionated style (e.g. with a 55%-gray 'Done' button), or an extension like GTMAppAuth wants to offer a Google flavored default config. Offering these default styles in the API as a factory implementation seems nicer to me than offering them as a block.  I expect 99% of developers will use one of the provided styles, making the "convenience" advantage of the block option a moot point.

So I'm casting my vote for 3). I'm guessing Steve is too since he proposed it. What do you think?

I realise this has been open a long time! Let's look at resolving it ASAP now.

maxsz:
Thanks @WilliamDenniss for coming back to this. I'm also happy with option 3) and PR #34. So I'm closing this.

PR #30, opened 2016-06-30
First step to add support for macOS.

msarda:
This PR is a first step in adding support for macOS in AppAuth. It follows the proposal on the following discussion on github https://github.com/openid/AppAuth-iOS/issues/20

This PR adds an authorization UI coordinator that is responsible to present the authorization UI. The iOS implementation continues to present a Safari View Controller or to open the Safari app is Safari View Controller is not supported on that iOS version.

This PR basically allows a user to provide a custom implementation of a authorization UI coordinator. Clients may thus provide a custom implementation that follows the guidelines in https://developers.google.com/identity/protocols/OAuth2InstalledApp for a mac that would take the user to the default browser and do the authorization flow.

TESTED=Locally using the Sample iOS application.

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/30)


WilliamDenniss:
Regarding the Podspec, can iOS and OSX targets co-exist within the a single XCode project?

If not, would it make more sense to have "AppAuth-iOS" and "AppAuth-OSX" as two separate fully contained pods (rather than with the "Core" pod dependency)?

WilliamDenniss:
Reviewed 15 of 15 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

msarda:
I am new to podspec (I have never used them before), however it looks like it works to have a single podspec per platform type (see https://github.com/core-plot/core-plot/blob/master/CorePlot.podspec for an example). I have updated the podspec description for AppAuth to cover this.

Note that my plan to add a macOS example app is to create a folder Example-Mac and to do a sample Mac application in there. I think that would be too much for the scope of this PR though (I have tried it locally and it works fine).

---

Review status: 13 of 14 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KLaq3-m7ZXLitozVRLQ:-2122158857)*

WilliamDenniss:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KLcM7yZI0NUOwhFwSov:1417088721)*

WilliamDenniss:
That looks much better!

LGTM

nubbel:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[AppAuth.podspec, line 30 \[r3\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KLex-1kA65tRsht3p3a:-KLex-1lwytYKquSE0me:2005500397) ([raw file](https://github.com/openid/appauth-ios/blob/a590860a7ee3f11edf4a3f8bd46b06211568b39e/AppAuth.podspec#L30)):*
> ```Ruby
>   s.requires_arc = true
> 
>   s.ios.exclude_files = "Source/**/*Mac.{h,m}"
> ```

Instead of excluding files for specific platforms, I would rather specify explicitly which files do belong to which platform using `s.{ios,osx}.source_files`.
Otherwise, when support for a new platform (e.g. tvOS or watchOS) was added, the exclude pattern would end up really complicated.

---

*[Source/OIDAuthorizationUICoordinatorIOS.h, line 29 \[r3\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KLezIGT_Td70YZkhYiP:-KLezIGT_Td70YZkhYiQ:2073329140) ([raw file](https://github.com/openid/appauth-ios/blob/a590860a7ee3f11edf4a3f8bd46b06211568b39e/Source/OIDAuthorizationUICoordinatorIOS.h#L29)):*
> ```objc
>         present an authorization request.
>  */
> @interface OIDAuthorizationUICoordinatorIOS : NSObject<OIDAuthorizationUICoordinator>
> ```

I think it is not necessary to include the platform in the class name. There will never be both a `OIDAuthorizationUICoordinatorIOS` and a `OIDAuthorizationUICoordinatorMac` in one target, hence they don't need differing names.
Alternatively, differentiate the coordinator names in terms of the strategy they implement (i.e. Safari view, (unsafe) web view, browser). That would make the library more extensible, allowing the user to choose a coordinator or create a custom one.
What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

WilliamDenniss:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[Source/OIDAuthorizationUICoordinatorIOS.h, line 29 \[r3\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KLezIGT_Td70YZkhYiP:-KLnGCsbLEZoyW6u0Zzt:-2063302677) ([raw file](https://github.com/openid/appauth-ios/blob/a590860a7ee3f11edf4a3f8bd46b06211568b39e/Source/OIDAuthorizationUICoordinatorIOS.h#L29)):*
<details><summary><i>Previously, nubbel (Dominique d'Argent) wrote…</i></summary>
> I think it is not necessary to include the platform in the class name. There will never be both a `OIDAuthorizationUICoordinatorIOS` and a `OIDAuthorizationUICoordinatorMac` in one target, hence they don't need differing names.
> Alternatively, differentiate the coordinator names in terms of the strategy they implement (i.e. Safari view, (unsafe) web view, browser). That would make the library more extensible, allowing the user to choose a coordinator or create a custom one.
> What do you think?
</details>
I agree that they don't *have* to be named after the platform, but I think they should definitely be different to avoid confusion, even if they won't ever be in the same target. 

I don't actually mind the "IOS" name here, you could say this represents the best practice, opinionated strategy for iOS.  It wouldn't stop someone to introduce another strategy of their choice, named accordingly, but the "IOS" one could remain our opinionated default strategy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

msarda:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[AppAuth.podspec, line 30 \[r3\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KLex-1kA65tRsht3p3a:-KLosV_fof3XAHhzC4KA:-499146009) ([raw file](https://github.com/openid/appauth-ios/blob/a590860a7ee3f11edf4a3f8bd46b06211568b39e/AppAuth.podspec#L30)):*
<details><summary><i>Previously, nubbel (Dominique d'Argent) wrote…</i></summary>
> Instead of excluding files for specific platforms, I would rather specify explicitly which files do belong to which platform using `s.{ios,osx}.source_files`.
> Otherwise, when support for a new platform (e.g. tvOS or watchOS) was added, the exclude pattern would end up really complicated.
</details>
Done.

---

*[Source/OIDAuthorizationUICoordinatorIOS.h, line 29 \[r3\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KLezIGT_Td70YZkhYiP:-KLosZmNXycLbvmDQITi:-425271309) ([raw file](https://github.com/openid/appauth-ios/blob/a590860a7ee3f11edf4a3f8bd46b06211568b39e/Source/OIDAuthorizationUICoordinatorIOS.h#L29)):*
<details><summary><i>Previously, WilliamDenniss (William Denniss) wrote…</i></summary>
> I agree that they don't *have* to be named after the platform, but I think they should definitely be different to avoid confusion, even if they won't ever be in the same target. 
> 
> I don't actually mind the "IOS" name here, you could say this represents the best practice, opinionated strategy for iOS.  It wouldn't stop someone to introduce another strategy of their choice, named accordingly, but the "IOS" one could remain our opinionated default strategy.
> 
> 
</details>
Keeping as is per William's comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

msarda:
My last POD config does not work as it excludes iOS and Mac specific files from all platforms. I'm working on fixing it.

---

Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KLpp-JHwe7WOHzPS8Ct:2110099655)*

msarda:
I decided to create per-OS source folders. This is needed to be able to included the files for each OS in the pods.

This PR is now ready for review. Please take another look.

---

Review status: 6 of 16 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KLpwfrnGWD8czXhB9-U:285124900)*

StevenEWright:
Review status: 6 of 16 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[Source/iOS/OIDAuthorizationService+IOS.h, line 38 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0uqEX4Eu8mtB-J4_S:-KM0uqEY_GQN39yP9guF:-1156274965) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthorizationService+IOS.h#L38)):*
> ```objc
>  */
> + (id<OIDAuthorizationFlowSession>)presentAuthorizationRequest:(OIDAuthorizationRequest *)request
>                                       presentingViewController:
> ```

Nit Monster: In this case we usually make following parameters left-aligned with the longest name starting in the 5th column (indented 4 spaces.) 

---

*[Source/iOS/OIDAuthorizationUICoordinatorIOS.m, line 89 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0vpgJS7sBneNEHsep:-KM0vpgJS7sBneNEHseq:1676885860) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthorizationUICoordinatorIOS.m#L89)):*
> ```objc
> }
> 
> - (void)cleanUp {
> ```

A comment stating that the weak references are being set to nil to avoid accidentally using them while none in an authorization flow may be warranted since it's not immediately apparent why these weak references are being set to nil here.

---

*[Source/iOS/OIDAuthState+IOS.h, line 42 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0wp75UuebJXofPwjb:-KM0wp76LWiqiHcL0Sa5:-1587373238) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthState+IOS.h#L42)):*
> ```objc
>                      presentingViewController:(UIViewController *)presentingViewController
>                                      callback:(OIDAuthStateAuthorizationCallback)callback;
> @end
> ```

Nit: In the other files we have generally left a blank line before `@end`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

StevenEWright:
Reviewed 6 of 21 files at r1, 14 of 16 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

StevenEWright:
LGTM

---

Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KM0xxfAr6RtWT1n-eTu:676834233)*

WilliamDenniss:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[Example/Example.xcodeproj/project.pbxproj, line 217 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM1Hvf7UVcpBF0YQ-Rs:-KM1Hvf7UVcpBF0YQ-Rt:59165405) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Example/Example.xcodeproj/project.pbxproj#L217)):*
> ```
> 			runOnlyForDeploymentPostprocessing = 0;
> 			shellPath = /bin/sh;
> 			shellScript = "diff \"${PODS_ROOT}/../Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [[ $? != 0 ]] ; then\n    cat << EOM\nerror: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\nEOM\n    exit 1\nfi\n";
> ```

This line is causing our CI process to fail – we don't have a Podfile for the Example app.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

WilliamDenniss:
Looked into the CI issue further. Just revert the changes to:
`Example/Example.xcodeproj/project.pbxproj`
`Example/Example.xcworkspace/contents.xcworkspacedata`

and the Example will build again. 

Looks like it was configured with pods – but our included Example doesn't need Pods to build.

msarda:
Review status: 14 of 15 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[Example/Example.xcodeproj/project.pbxproj, line 217 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM1Hvf7UVcpBF0YQ-Rs:-KM3h_-oeEWnPpmmprwZ:-1702159038) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Example/Example.xcodeproj/project.pbxproj#L217)):*
<details><summary><i>Previously, WilliamDenniss (William Denniss) wrote…</i></summary>
> This line is causing our CI process to fail – we don't have a Podfile for the Example app.
</details>
I revert the changes to the Example xcodeproj and xcworkspace files.

Note that I've had to add a linker flag "-ObjC" to the Example project as it needs to link symbols that are defined in categories and without this flag they are ignored when the project is linked. I think that this is not an issue when using CocoaPods to configure the project, as it seems that this flag is automatically set when the pod is installed.

For more info about this flag, see http://stackoverflow.com/questions/6629979/what-does-the-objc-linker-flag-do

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

msarda:
Review status: 9 of 15 files reviewed at latest revision, 6 unresolved discussions.

---

*[Source/iOS/OIDAuthorizationService+IOS.h, line 38 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0uqEX4Eu8mtB-J4_S:-KM3kgqF2TsbDIcLgAmA:-499146009) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthorizationService+IOS.h#L38)):*
<details><summary><i>Previously, StevenEWright (Steven E Wright) wrote…</i></summary>
> Nit Monster: In this case we usually make following parameters left-aligned with the longest name starting in the 5th column (indented 4 spaces.) 
</details>
Done.

---

*[Source/iOS/OIDAuthorizationUICoordinatorIOS.m, line 89 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0vpgJS7sBneNEHsep:-KM3kdaI3IUK1NRpUKI7:-499146009) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthorizationUICoordinatorIOS.m#L89)):*
<details><summary><i>Previously, StevenEWright (Steven E Wright) wrote…</i></summary>
> A comment stating that the weak references are being set to nil to avoid accidentally using them while none in an authorization flow may be warranted since it's not immediately apparent why these weak references are being set to nil here.
</details>
Done.

---

*[Source/iOS/OIDAuthState+IOS.h, line 42 \[r6\]](https://reviewable.io:443/reviews/openid/appauth-ios/30#-KM0wp75UuebJXofPwjb:-KM3keiUhCKUzPM9qfdR:-499146009) ([raw file](https://github.com/openid/appauth-ios/blob/e6115b1b478dfeb63bf8e58bcd977f450483ffe2/Source/iOS/OIDAuthState+IOS.h#L42)):*
<details><summary><i>Previously, StevenEWright (Steven E Wright) wrote…</i></summary>
> Nit: In the other files we have generally left a blank line before `@end`.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

WilliamDenniss:
Reviewed 9 of 16 files at r5, 1 of 2 files at r7, 5 of 5 files at r8.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

WilliamDenniss:
LGTM

StevenEWright:
:lgtm:

---

Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30#-:-KM5wa6XAEabF8A602rE:1417088721)*

StevenEWright:
Reviewed 4 of 16 files at r5, 1 of 2 files at r7, 5 of 5 files at r8.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/30)*

PR #33, opened 2016-07-07
Nonnull initializers

nubbel:
This change removes `nullable` annotations from user-facing initializers which makes them much nicer to use at the Swift call-site:

_Before:_
```swift
let config = OIDServiceConfiguration(authorizationEndpoint: authEndpoint,
		                                     tokenEndpoint: tokenEndpoint)!
```

_After (no force-unwrapping required):_
```swift
let config = OIDServiceConfiguration(authorizationEndpoint: authEndpoint,
		                                     tokenEndpoint: tokenEndpoint)
```

I've checked the initializers if there is any chance they could ever return `nil` and it turned out there is none. They do call `NSObject`'s `init` which guarantees to never return `nil`.

There are many more initializers in this project marked as `nullable`, but these two classes are the most "visible" ones to library users.

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/33)

### DIFF
--- a/Source/OIDAuthorizationRequest.h
+++ b/Source/OIDAuthorizationRequest.h
@@ -127,7 +127,7 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
     @brief Unavailable. Please use
         @c initWithConfiguration:clientId:scopes:redirectURL:additionalParameters:.
  */
-- (nullable instancetype)init NS_UNAVAILABLE;
+- (instancetype)init NS_UNAVAILABLE;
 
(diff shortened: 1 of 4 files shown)


WilliamDenniss:
Is it worth reviewing the other initializers?

:lgtm: 

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/33#-:-KM5srvmowJSKnLZJ371:243883729)*

nubbel:
Cool, thanks 🍻

I've spotted some initializers that can actually return nil, but the majority doesn't. It might be worth going over them and remove `nullable` where possible.

PR #39, opened 2016-07-27
Access token and id token can be nil when OIDAuthStateAction is called.

jeromelebel:
Access token and id token can be nil when OIDAuthStateAction is called (when having an error). The definition of OIDAuthStateAction is correct, but the callback is wrongly defined in AppAuthExampleViewController.m.

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/39)


WilliamDenniss:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/39#-:-KNeVvNmkBWlMOzHKliv:1417088721)*

PR #51, opened 2016-11-01
ARC-ified the loopback HTTP server class.

WilliamDenniss:
Notes:
– `HTTPConnection` objects were previously self-retaining. HTTPServer now keeps track of them instead so they are retained correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/51)

### DIFF
--- a/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
+++ b/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
@@ -24,12 +24,10 @@
 @implementation HTTPServer
 
 - (id)init {
+    self = [super init];
     connClass = [HTTPConnection self];
+    connections = [[NSMutableArray alloc] init];
     return self;
-}
-
-- (void)dealloc {
-    [super dealloc];
 }
 
 - (Class)connectionClass {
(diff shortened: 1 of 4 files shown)


Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
```diff
@@ -24,12 +24,13 @@
 @implementation HTTPServer
 
 - (id)init {
+    self = [super init];
     connClass = [HTTPConnection self];
+    connections = [[NSMutableArray alloc] init];
     return self;
 }
 
 - (void)dealloc {
```
msarda: Remove the dealloc function (it does not do anything now).

Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
```diff
@@ -95,8 +93,7 @@ - (id)initWithPeerAddress:(NSData *)addr inputStream:(NSInputStream *)istr outpu
...
     [self invalidate];
-    [peerAddress release];
-    [super dealloc];
+    peerAddress = nil;
```
msarda: Optional nit (here and everywhere else) I think setting ivars to nil in a dealloc is not really needed.

WilliamDenniss: OK, I cleared up the ones I introduced.

PR #55, opened 2016-11-06
Documentation and API improvements.

WilliamDenniss:
– `withFreshTokensPerformAction` renamed based on feedback
– Syncronized the public API exposed in AppAuth.h and documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/55)


Source/OIDTokenRequest.h
```diff
@@ -113,7 +113,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable instancetype)init NS_UNAVAILABLE;
 
-/*! @fn initWithConfiguration:grantType:authorizationCode:redirectURL:clientID:scopes:refreshToken:codeVerifier:additionalParameters:
+/*! @fn initWithConfiguration:grantType:authorizationCode:redirectURL:clientID:clientSecret:scopes:refreshToken:codeVerifier:additionalParameters:
```
msarda: Based on the explanation on Doxygen, fn should be omitted when used in front of a function (https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdfn).

Bared in mind that I have never use Doxygen though (however this is what you're doing in other files).

PR #59, opened 2016-11-18
Correctly returns access_denied error when the user declines.

### LINKED ISSUE #56: If user clicks on deny error code is incorrect
If user clicked on deny expected code is ```OIDErrorCodeOAuthAccessDenied```, but really have this ```OIDErrorCodeOAuthAuthorizationClientError```, as this condition always true for this case, and it changes source error:

```
  if (!OIDIsEqualIncludingNil(_request.state, response.state)) {
    NSMutableDictionary *userInfo = [query.dictionaryValue mutableCopy];
    userInfo[NSLocalizedFailureReasonErrorKey] =
        [NSString stringWithFormat:@"State mismatch, expecting %@ but got %@ in authorization "
                                    "response %@",
                                   _request.state,
                                   response.state,
                                   response];
    response = nil;
    error = [NSError errorWithDomain:OIDOAuthAuthorizationErrorDomain
                                code:OIDErrorCodeOAuthAuthorizationClientError
                            userInfo:userInfo];
  }
```

WilliamDenniss:
– Fixes #56

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/59)

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -34,6 +34,10 @@
  */
 static NSString *const kOpenIDConfigurationWellKnownPath = @".well-known/openid-configuration";
 
+/*! @brief The state authorization parameter.
+ */
+static NSString *const kStateParameter = @"state";
+
 NS_ASSUME_NONNULL_BEGIN
 
 @interface OIDAuthorizationFlowSessionImplementation : NSObject<OIDAuthorizationFlowSession>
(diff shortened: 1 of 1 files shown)


StevenEWright:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Source/OIDAuthorizationService.m, line 127 at r1](https://reviewable.io:443/reviews/openid/appauth-ios/59#-KWqHn4GOfLjTeYcx8WM:-KWqHn4GOfLjTeYcx8WN:bww2t7e) ([raw file](https://github.com/openid/appauth-ios/blob/f1e0ee8cb6118008056b0312237d862d56535574/Source/OIDAuthorizationService.m#L127)):*
> ```objc
> 
>   // verifies that the state in the response matches the state in the request, or both are nil
>   if (!OIDIsEqualIncludingNil(_request.state, query.dictionaryValue[kStateParameter])) {
> ```

It's unfortunate we can't use the existing code in OIDAuthorizationResponse to extract the state parameter... just seems like it would have been cleaner that way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/59#-:-KWqIAxkNYsyWpPDWsUA:bnfp4nl)*

WilliamDenniss:
I agree. The way AppAuth has the `*Response` classes setup though, they only ever represent a successful response.

PR #65, opened 2016-12-04
Sets the NSError localizedDescription in more cases.

WilliamDenniss:
I heard some reports that it was hard to diagnose issues in AppAuth. While testing #22 I noticed that they were fairly opaque.

I believe the issue is that `localizedDescription` will contain a default description if we don't set one.  We were setting a `localizedFailureReason`, but when the developer calls `localizedDescription` and they get a vague default error, they may not think to check `localizedFailureReason`.

This PR sets localizedDescription overriding the vague default, and leaves localizedFailureReason nil. In the future if we were to set localizedFailureReason, we should always ensure that localizedDescription is included to avoid this usability issue.

Also passed through the description for the non-OAuth errors (like network errors).

---
This change is [Reviewable](https://reviewable.io/reviews/openid/appauth-ios/65)

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -126,7 +126,7 @@ NS_ASSUME_NONNULL_BEGIN
   // verifies that the state in the response matches the state in the request, or both are nil
   if (!OIDIsEqualIncludingNil(_request.state, query.dictionaryValue[kStateParameter])) {
     NSMutableDictionary *userInfo = [query.dictionaryValue mutableCopy];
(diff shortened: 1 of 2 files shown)


StevenEWright:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openid/appauth-ios/65#-:-KY6suh387SP3SI1dA-j:bnfp4nl)*

PR #100, opened 2017-03-25
Support for Basic Client Authentication

WilliamDenniss:
With this change, AppAuth will support no-client-auth (suitable for clients without a secret) and Basic client auth (suitable for clients with a secret).

Previously AppAuth supported no-client-auth, and client credentials in the request body.

Basic auth support is [required](https://tools.ietf.org/html/rfc6749#section-2.3.1) by OAuth 2.0, which also recommends against client credentials in the request body when the client is capable of doing Basic auth (which AppAuth is capable of).

Replaces #24, but doesn't offer the extensibility that #24 did.


ve7jtb:
LGTM

Issue #4, opened 2016-03-07
Support "nonce" OpenID Connect auth request parameter


iainmcgin:
[OpenID Connect Core 1.0, Section 3.1.2.1](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest), "nonce" parameter:

> OPTIONAL. String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. Sufficient entropy MUST be present in the nonce values used to prevent attackers from guessing values. For implementation notes, see [Section 15.5.2](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes).

[Section 3.2.2.1](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest), "Authentication Request" specifies that this parameter is REQUIRED for implicit flow authentication requests (those with response type `id_token` or `id_token token`).

WilliamDenniss:
This has been implemented.

Issue #17, opened 2016-03-07
Support JWT decoding and validation


iainmcgin:
Support validating JWTs and extracting their claims as a map. This will require the ability to either dynamically use the `jwks_uri` keys provided by the provider's [discovery document](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata), or a set of acceptable keys provided by the developer.

WilliamDenniss:
This was implemented.

Issue #12, opened 2016-03-07
Support "claims" OpenID Connect auth request parameter


iainmcgin:
[OpenID Connect Core 1.0, Section 5.5](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter), `claims` parameter:

> OPTIONAL. This parameter is used to request that specific Claims be returned. The value is a JSON object listing the requested Claims.
> 
> An example Claims request is as follows:
> 
> ``` json
> {
>   "userinfo":
>   {
>       "given_name": {"essential": true},
>       "nickname": null,
>       "email": {"essential": true},
>       "email_verified": {"essential": true},
>       "picture": null,
>       "http://example.info/claims/groups": null
>   },
>   "id_token":
>   {
>       "auth_time": {"essential": true},
>       "acr": {"values": ["urn:mace:incommon:iap:silver"] }
>   }
> }
> ```

WilliamDenniss:
Note that this is supported via the `additionalParameters` param in the `OIDAuthorizationRequest` initializers.

alvincrisuy:
How to use this? as it only accepts NSDictionary<NSString *, NSString *> *)

ShimelesT:
what is the exact purpose of value? can a client pass values to OP? 
"acr": {"values": ["urn:mace:incommon:iap:silver"] } 
or is it for verification?

WilliamDenniss:
> How to use this? as it only accepts NSDictionary<NSString *, NSString *> *)

In this case, "claims" is the NSString key, and the JSON-encoded NSString is the value.

additionalParameters are simply added to the URL in key=value format. OpenID is expecting the value to be a JSON-encoded string, so that's what you should supply.

WilliamDenniss:
> what is the exact purpose of value? can a client pass values to OP?
> "acr": {"values": ["urn:mace:incommon:iap:silver"] }
> or is it for verification?

It allows you to request from the OP certain claims that you want included in the ID Token. Here's the text from the [spec](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter):

> The claims Authentication Request parameter requests that specific Claims be returned from the UserInfo Endpoint and/or in the ID Token. It is represented as a JSON object containing lists of Claims being requested from these locations. Properties of the Claims being requested MAY also be specified.

Actually delivering those claims is really OP-dependent, so probably best to check with them what claims can be requested in this way.

v-khatri:
[...]

PR #101, opened 2017-03-25
ID Token parsing and OpenID RP Certification

WilliamDenniss:
## ID Token Parsing
* The 5 required fields are exposed as params.
* Authorization code exchange exchange now validates the ID Token iss, aud, and iat claims.

### Response Type and Response Mode
- [x] rp-response_type-code
### scope Request Parameter
- [x] rp-scope-userinfo-claims
### nonce Request Parameter
- [x] rp-nonce-invalid
### Client Authentication
- [x] rp-token_endpoint-client_secret_basic
### ID Token
- [x] rp-id_token-aud
- [ ] rp-id_token-kid-absent-single-jwks*
- [x] rp-id_token-sig-none
- [x] rp-id_token-issuer-mismatch
- [ ] rp-id_token-kid-absent-multiple-jwks*
- [x] rp-id_token-bad-sig-rs256*
- [x] rp-id_token-iat	
- [x] rp-id_token-sig-rs256*
- [x] rp-id_token-sub	

*AppAuth for iOS & macOS only supports `alg:none` validation.

### UserInfo Endpoint
- [x] rp-userinfo-bad-sub-claim
- [x] rp-userinfo-bearer-header

## Fixes:
#4  Support "nonce" OpenID Connect auth request parameter.
#17 Support JWT decoding and validation (conditional: alg:none based validation only).
#102 OpenID Connect RP Conformance Testing.

### DIFF
--- a/Source/OIDAuthState.m
+++ b/Source/OIDAuthState.m
@@ -125,7 +125,8 @@ static const NSUInteger kExpiryTimeTolerance = 60;
                                OIDTokenRequest *tokenExchangeRequest =
                                    [authorizationResponse tokenExchangeRequest];
                                [OIDAuthorizationService
-                                   performTokenRequest:tokenExchangeRequest
+                                             performTokenRequest:tokenExchangeRequest
+                                   originalAuthorizationResponse:authorizationResponse
                                               callback:^(OIDTokenResponse *_Nullable tokenResponse,
                                                          NSError *_Nullable tokenError) {
                                                 OIDAuthState *authState;
@@ -473,6 +474,7 @@ static const NSUInteger kExpiryTimeTolerance = 60;
   OIDTokenRequest *tokenRefreshRequest =
       [self tokenRefreshRequestWithAdditionalParameters:additionalParameters];
   [OIDAuthorizationService performTokenRequest:tokenRefreshRequest
+                 originalAuthorizationResponse:_lastAuthorizationResponse
                                       callback:^(OIDTokenResponse *_Nullable response,
                                                  NSError *_Nullable error) {
     dispatch_async(dispatch_get_main_queue(), ^() {
--- a/Source/OIDAuthState.m
+++ b/Source/OIDAuthState.m
@@ -124,10 +124,9 @@ static const NSUInteger kExpiryTimeTolerance = 60;
                                // code exchange
                                OIDTokenRequest *tokenExchangeRequest =
                                    [authorizationResponse tokenExchangeRequest];
-                               [OIDAuthorizationService
-                                             performTokenRequest:tokenExchangeRequest
-                                   originalAuthorizationResponse:authorizationResponse
-                                              callback:^(OIDTokenResponse *_Nullable tokenResponse,
+                               [OIDAuthorizationService performTokenRequest:tokenExchangeRequest
+                                              originalAuthorizationResponse:authorizationResponse
+                                   callback:^(OIDTokenResponse *_Nullable tokenResponse,
                                                          NSError *_Nullable tokenError) {
                                                 OIDAuthState *authState;
                                                 if (tokenResponse) {
@@ -137,7 +136,7 @@ static const NSUInteger kExpiryTimeTolerance = 60;
                                                                       tokenResponse:tokenResponse];
                                                 }
                                                 callback(authState, tokenError);
-                                              }];
+                               }];
                              } else {
                                // implicit or hybrid flow (hybrid flow assumes code is not for this
                                // client)
--- a/Source/OIDAuthorizationRequest.h
+++ b/Source/OIDAuthorizationRequest.h
@@ -47,6 +47,7 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
   NSString *_scope;
   NSURL *_redirectURL;
   NSString *_state;
+  NSString *_nonce;
   NSString *_codeVerifier;
   NSString *_codeChallenge;
   NSString *_codeChallengeMethod;
@@ -108,6 +109,17 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
  */
 @property(nonatomic, readonly, nullable) NSString *state;
 
+/*! @brief String value used to associate a Client session with an ID Token, and to mitigate replay
+        attacks. The value is passed through unmodified from the Authentication Request to the ID
+        Token. Sufficient entropy MUST be present in the nonce values used to prevent attackers from
+        guessing values.
+    @remarks nonce
+    @discussion If this value is not explicitly set, this library will automatically add nonce and
+        perform appropriate validation of the nonce in the ID Token.
+    @see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+ */
+@property(nonatomic, readonly, nullable) NSString *nonce;
+
 /*! @brief The PKCE code verifier.
     @remarks code_verifier
     @discussion The code verifier itself is not included in the authorization request that is sent
@@ -160,8 +172,8 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
              responseType:(NSString *)responseType
      additionalParameters:(nullable NSDictionary<NSString *, NSString *> *)additionalParameters;
 
-/*! @brief Creates an authorization request with opinionated defaults (a secure @c state, and
-        PKCE with S256 as the @c code_challenge_method).
+/*! @brief Creates an authorization request with opinionated defaults (a secure @c state, @c nonce,
+        and PKCE with S256 as the @c code_challenge_method).
     @param configuration The service's configuration.
     @param clientID The client identifier.
     @param clientSecret The client secret.
@@ -181,7 +193,7 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
              responseType:(NSString *)responseType
      additionalParameters:(nullable NSDictionary<NSString *, NSString *> *)additionalParameters;
 
-/*! @brief Designated initializer.
+/*! @brief Deprecated, replaced with @c OIDAuthState.initWithConfiguration:clientId:clientSecret:scope:redirectURL:responseType:state:nonce:codeVerifier:codeChallenge:codeChallengeMethod:additionalParameters:.
     @param configuration The service's configuration.
     @param clientID The client identifier.
     @param scope A scope string per the OAuth2 spec (a space-delimited set of scopes).
@@ -206,6 +218,39 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
               redirectURL:(nullable NSURL *)redirectURL
              responseType:(NSString *)responseType
                     state:(nullable NSString *)state
+             codeVerifier:(nullable NSString *)codeVerifier
+            codeChallenge:(nullable NSString *)codeChallenge
+      codeChallengeMethod:(nullable NSString *)codeChallengeMethod
+     additionalParameters:(nullable NSDictionary<NSString *, NSString *> *)additionalParameters
+__deprecated_msg("Replaced with OIDAuthState.initWithConfiguration:clientId:clientSecret:scope:redirectURL:responseType:state:nonce:codeVerifier:codeChallenge:codeChallengeMethod:additionalParameters:");
+
+/*! @brief Designated initializer.
+    @param configuration The service's configuration.
+    @param clientID The client identifier.
+    @param scope A scope string per the OAuth2 spec (a space-delimited set of scopes).
+    @param redirectURL The client's redirect URI.
+    @param responseType The expected response type.
+    @param state An opaque value used by the client to maintain state between the request and
+        callback.
+    @param nonce String value used to associate a Client session with an ID Token.
+    @param codeVerifier The PKCE code verifier. See @c OIDAuthorizationRequest.generateCodeVerifier.
+    @param codeChallenge The PKCE code challenge, calculated from the code verifier such as with
+        @c OIDAuthorizationRequest.codeChallengeS256ForVerifier:.
+    @param codeChallengeMethod The PKCE code challenge method.
+        ::OIDOAuthorizationRequestCodeChallengeMethodS256 when
+        @c OIDAuthorizationRequest.codeChallengeS256ForVerifier: is used to create the code
+        challenge.
+    @param additionalParameters The client's additional authorization parameters.
+ */
+- (instancetype)
+    initWithConfiguration:(OIDServiceConfiguration *)configuration
+                 clientId:(NSString *)clientID
+             clientSecret:(nullable NSString *)clientSecret
+                    scope:(nullable NSString *)scope
+              redirectURL:(nullable NSURL *)redirectURL
+             responseType:(NSString *)responseType
+                    state:(nullable NSString *)state
+                    nonce:(nullable NSString *)nonce
              codeVerifier:(nullable NSString *)codeVerifier
             codeChallenge:(nullable NSString *)codeChallenge
       codeChallengeMethod:(nullable NSString *)codeChallengeMethod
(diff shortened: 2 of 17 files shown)


Source/OIDAuthorizationService.m
```diff
@@ -344,6 +345,57 @@ + (void)performTokenRequest:(OIDTokenRequest *)request callback:(OIDTokenCallbac
...
+          [OIDErrorUtilities errorWithCode:OIDErrorCodeIDTokenParsingError
+                           underlyingError:nil
+                               description:@"ID Token parsing failed"];
+        dispatch_async(dispatch_get_main_queue(), ^{
+          callback(nil, invalidIDToken);
+        });
+        return;
+      }
+
+      NSURL *issuer = tokenResponse.request.configuration.discoveryDocument.issuer;
```
ve7jtb: The issuer always needs to be validated.   If not doing discovery then the issuer would need to be statically along with Authorization endpoint etc.

WilliamDenniss: Added the ability to manually specify issuer when not using discovery.

zboralski:
any update?

zboralski:
also any reason why an implicit flow with a simple 'id_token' response type shouldn't be supported?

WilliamDenniss:
Updated the documentation & code comments for the ID Token features. Should be good to go now.

Here is a good summary (taken from the code) of the validation that is performed. Basically we are following the OpenID Connect [specification for verification of ID Tokens](http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) on the code flow, with the important note that we are exercising the option to rely only on the TLS validation of the direct connection between the token endpoint and client, and are not performing JWT signature verification (users of AppAuth are welcome to add this themselves, should they wish).

    // If an ID Token is included in the response, validates the ID Token following the rules
    // in OpenID Connect Core Section 3.1.3.7 for features that AppAuth directly supports
    // (which excludes rules #1, #4, #5, #7, #8, #12, and #13). Regarding rule #6, ID Tokens
    // received by this class are received via direct communication between the Client and the Token
    // Endpoint, thus we are exercising the option to rely only on the TLS validation. AppAuth
    // has a zero dependencies policy, and verifying the JWT signature would add a dependency.
    // Users of the library are welcome to perform the JWT signature verification themselves should
    // they wish.

jakeholland:
@WilliamDenniss - Need anything else for this? I have a project in flight this would be great for.

zboralski:
Any progress on this? Do you need any help?

zboralski:
@WilliamDenniss @StevenEWright What's the status of this pull request?

StevenEWright:
@zboralski thank you for the ping. Ugh. Looks like I really dropped the ball on this. I’m sorry. I will look tonight.

Source/OIDIDToken.m
```diff
@@ -0,0 +1,152 @@
...
+        distributed under the License is distributed on an "AS IS" BASIS,
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+ */
+
+#import "OIDIDToken.h"
+
+/*! Field keys associated with an ID Token. */
+static NSString *const kIssKey = @"iss";
```
StevenEWright: Any reason not to decode acr, amr, azp as well if present?

Source/OIDIDToken.m
```diff
@@ -0,0 +1,152 @@
...
+@synthesize subject = _subject;
+@synthesize audience = _audience;
+@synthesize expiresAt = _expiresAt;
+@synthesize issuedAt = _issuedAt;
+@synthesize nonce = _nonce;
+
+- (instancetype)initWithIDTokenString:(NSString *)idToken {
+  self = [super init];
+  NSArray *sections = [idToken componentsSeparatedByString:@"."];
+  if (sections.count > 1) {
```
StevenEWright: I did a double-take here at the end of the method. Might be nice to:

```
// The header and claims are required.
if (sections.count <= 1) {
  return nil;
}
```

Source/OIDIDToken.m
```diff
@@ -0,0 +1,152 @@
...
+    fieldMap[kSubKey] =
+        [[OIDFieldMapping alloc] initWithName:@"_subject" type:[NSString class]];
+    fieldMap[kAudKey] =
+        [[OIDFieldMapping alloc] initWithName:@"_audience"
+                                         type:[NSArray class]
+                                   conversion:^id _Nullable(NSObject *_Nullable value) {
+          if ([value isKindOfClass:[NSArray class]]) {
+            return value;
+          }
+          return @[value];
```
StevenEWright: In this case it's expected value is a type of string, right? You've been so diligent about checking types... do we want to check that as well?

Source/OIDIDToken.m
```diff
@@ -0,0 +1,152 @@
...
+        }];
+    fieldMap[kExpKey] =
+        [[OIDFieldMapping alloc] initWithName:@"_expiresAt"
+                                         type:[NSDate class]
+                                   conversion:^id _Nullable(NSObject *_Nullable value) {
+          if (![value isKindOfClass:[NSNumber class]]) {
+            return value;
+          }
+          NSNumber *valueAsNumber = (NSNumber *)value;
+          return [NSDate dateWithTimeIntervalSince1970:[valueAsNumber longLongValue]];
```
StevenEWright: dot notation for properties? (longLongValue below as well)

Source/OIDServiceConfiguration.h
```diff
@@ -48,6 +49,10 @@ typedef void (^OIDServiceConfigurationCreated)
  */
 @property(nonatomic, readonly) NSURL *tokenEndpoint;
 
+/*! @brief The token exchange and refresh endpoint URI.
+ */
+@property(nonatomic, readonly) NSURL *issuer;
```
StevenEWright: Should this be nullable?

```
@property(nonatomic, readonly, nullable) NSURL *issuer;
```

Source/OIDServiceConfiguration.h
```diff
@@ -76,6 +81,24 @@ typedef void (^OIDServiceConfigurationCreated)
                                 tokenEndpoint:(NSURL *)tokenEndpoint
                          registrationEndpoint:(nullable NSURL *)registrationEndpoint;
 
+/*! @param authorizationEndpoint The authorization endpoint URI.
+    @param tokenEndpoint The token exchange and refresh endpoint URI.
+    @param issuer The OpenID Connect issuer.
+ */
+- (instancetype)initWithAuthorizationEndpoint:(NSURL *)authorizationEndpoint
```
StevenEWright: Should we mark `init` as unavailable in the header?

Source/OIDServiceConfiguration.m
```diff
@@ -54,25 +59,27 @@ @implementation OIDServiceConfiguration
...
 
 - (instancetype)init
     OID_UNAVAILABLE_USE_INITIALIZER(@selector(
         initWithAuthorizationEndpoint:
-                       tokenEndpoint:
-                registrationEndpoint:)
+                       tokenEndpoint:)
     );
 
 - (instancetype)initWithAuthorizationEndpoint:(NSURL *)authorizationEndpoint
```
StevenEWright: Looks like this is the actual designated initializer? Should we mark it as such in the header?

Source/OIDAuthorizationService.m
```diff
@@ -348,6 +358,120 @@ + (void)performTokenRequest:(OIDTokenRequest *)request callback:(OIDTokenCallbac
...
+        dispatch_async(dispatch_get_main_queue(), ^{
+          callback(nil, invalidIDToken);
+        });
+        return;
+      }
+      
+      // OpenID Connect Core Section 3.1.3.7. rule #11
+      // Validates the nonce.
+      NSString *nonce = authorizationResponse.request.nonce;
+      if (nonce && ![idToken.nonce isEqual:nonce]) {
```
jhollandjimmyjohn: Not suer if you are already aware, but there is a problem here. A nonce isn't sent with the call and then it expects one here.

WilliamDenniss: If the authorization request doesn't set nonce, then this check will be skipped (`if (nonce &&`… should achieve that). Are you sure there's a problem?

jhollandjimmyjohn: I have a project using this library with an IdentityServer4 server. The problem is that `authorizationResponse.request.nonce` is a valid nonce and idToken.nonce is null.

WilliamDenniss: [OpenID Connect Core Section 2](http://openid.net/specs/openid-connect-core-1_0.html#IDToken) states:

> If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request. 

I believe we are implementing this correctly in AppAuth , is there a bug in IdentityServer4?

jhollandjimmyjohn: I have opened a bug with IdentityServer4 here: https://github.com/IdentityServer/IdentityServer4/issues/2180

jhollandjimmyjohn: [Section 12.1](http://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken)
Does not state anything about needing a nonce to verify a Refresh Request. Not sure AppAuth is really doing the right thing here.

WilliamDenniss: It's a bit ambiguous, but yeah I don't think the intention was for nonce to be required in the ID Token for the token response during refreshes. I sent a [note](http://lists.openid.net/pipermail/openid-specs-ab/Week-of-Mon-20180416/006728.html) to the Connect mailing list regarding the possible ambiguity. Google doesn't return the nonce in refreshed ID Tokens either.

I've fixed this in the PR. Thanks very much for engaging and reporting this problem!

Source/OIDIDToken.h
```diff
@@ -0,0 +1,101 @@
...
+    @remarks exp
+    @see http://openid.net/specs/openid-connect-core-1_0.html#IDToken
+ */
+@property(nonatomic, readonly) NSDate *expiresAt;
+
+/*! @brief Time at which the JWT was issued.
+    @remarks iat
+    @see http://openid.net/specs/openid-connect-core-1_0.html#IDToken
+ */
+@property(nonatomic, readonly) NSDate *issuedAt;
```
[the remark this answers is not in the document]
WilliamDenniss: Appears to be required to me:

> iat
> REQUIRED. Time at which the JWT was issued. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time.

I think you're quoting from `auth_time`?

Source/OIDAuthorizationRequest.h
```diff
@@ -46,6 +46,7 @@ extern NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256;
   NSString *_scope;
   NSURL *_redirectURL;
   NSString *_state;
+  NSString *_nonce;
```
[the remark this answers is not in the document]
WilliamDenniss: Hmm, at least the property itself is declared as nullable.  I hope to remove this code in the near future, once 32bit is gone.

Source/OIDIDToken.m
WilliamDenniss: None are used by the library directly, and everything is provided in the `claims` dictionary. I didn't decode all known types to reduce the scope. We *could* do it in the future, if there's value.

It's a bit like the other standard fields that we don't support of connect, and tell people just to use "additionalParams" for.

Source/OIDAuthorizationService.m
```diff
@@ -348,6 +358,120 @@ + (void)performTokenRequest:(OIDTokenRequest *)request callback:(OIDTokenCallbac
...
+                             description:@"ID Token expired"];
+        dispatch_async(dispatch_get_main_queue(), ^{
+          callback(nil, invalidIDToken);
+        });
+        return;
+      }
+      
+      // OpenID Connect Core Section 3.1.3.7. rule #10
+      // Validates that the issued at time is not more than +/- 5 minutes on the current time.
+      NSTimeInterval issuedAtDifference = [idToken.issuedAt timeIntervalSinceNow];
```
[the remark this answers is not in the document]
WilliamDenniss: I don't think too many people set the time manually, I agree – but perhaps some do. I'm more concerned about server clock skew.

I don't really know what const to use here. The docs say we should "reject tokens that were issued too far away from the current time", but don't specify what "too far away" is, so I plucked 5 minutes out of the air, since it seemed short, but also long enough to account for multiple-minutes of clock skew on both sides.

Source/OIDServiceConfiguration.h
WilliamDenniss: I think we do already? `- (instancetype)init NS_UNAVAILABLE;`

Source/OIDServiceConfiguration.m
WilliamDenniss: So this method is not actually exposed in the header at all.

We're expecting people to either specify the endpoints manually, OR use a discovery doc – but not both. Do you think we should revise this?

Source/OIDAuthState.m
```diff
@@ -125,7 +125,8 @@ @implementation OIDAuthState
                                OIDTokenRequest *tokenExchangeRequest =
                                    [authorizationResponse tokenExchangeRequest];
                                [OIDAuthorizationService
-                                   performTokenRequest:tokenExchangeRequest
+                                             performTokenRequest:tokenExchangeRequest
+                                   originalAuthorizationResponse:authorizationResponse
                                               callback:^(OIDTokenResponse *_Nullable tokenResponse,
```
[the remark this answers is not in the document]
WilliamDenniss: You'll be pleased to know we're still picky. I took another stab at it, this one is kind of hard to fit in, PTAL. If you have any guidance please let me know.

Source/OIDAuthorizationService.m
```diff
@@ -348,6 +358,120 @@ + (void)performTokenRequest:(OIDTokenRequest *)request callback:(OIDTokenCallbac
...
+
+      // OpenID Connect Core Section 3.1.3.7. rules #7 & #8
+      // Not applicable. See rule #6.
+
+      // OpenID Connect Core Section 3.1.3.7. rule #9
+      // Validates that the current time is before the expiry time.
+      NSTimeInterval expiresAtDifference = [idToken.expiresAt timeIntervalSinceNow];
+      if (expiresAtDifference < 0) {
+        NSError *invalidIDToken =
+        [OIDErrorUtilities errorWithCode:OIDErrorCodeIDTokenFailedValidationError
```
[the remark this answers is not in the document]
WilliamDenniss: Done

Source/OIDIDToken.h
```diff
@@ -0,0 +1,101 @@
...
+
+/*! @brief String value used to associate a Client session with an ID Token, and to mitigate replay
+        attacks.
+    @remarks nonce
+    @see http://openid.net/specs/openid-connect-core-1_0.html#IDToken
+ */
+@property(nonatomic, readonly) NSString *nonce;
```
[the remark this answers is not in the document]
WilliamDenniss: Yes. Done.

Source/OIDIDToken.m
WilliamDenniss: Done.

Source/OIDIDToken.m
WilliamDenniss: Done.

Source/OIDServiceConfiguration.h
WilliamDenniss: Yes. Done.

Source/OIDIDToken.m
WilliamDenniss: Done.

StevenEWright:
LGTM

ShaneQi:
Our identify server requires mobile clients to send nonce. I verified that this branch will work for our case. Are we merging this PR in the near future?

Thanks!

PR #106, opened 2017-04-10
Update OpenID Discovery to validate Authorization Endpoint is valid

brandwe:
I've read through the spec, and I think it is the authorization server that needs to be validated as resolvable and not the issuer. I don't see where it is required that the issuer == authorization server (but I could be wrong).

This is actually a problem for Microsoft, because we support multiple tenants and therefore discovery endpoints and issuers that are anonymous at first, particularly using the `/common`, `/orgaizations`, `/consumer` discovery endpoints. This results in correct authorization and token endpoints, but results in an issuer being a placeholder `{tenant id}`. 

As an example, here is the JSON returned from calling our `/common` endpoint:

```
{"authorization_endpoint":"https://login.microsoftonline.com/common/oauth2/v2.0/authorize","token_endpoint":"https://login.microsoftonline.com/common/oauth2/v2.0/token","token_endpoint_auth_methods_supported":["client_secret_post","private_key_jwt"],"jwks_uri":"https://login.microsoftonline.com/common/discovery/v2.0/keys","response_modes_supported":["query","fragment","form_post"],"subject_types_supported":["pairwise"],"id_token_signing_alg_values_supported":["RS256"],"http_logout_supported":true,"frontchannel_logout_supported":true,"end_session_endpoint":"https://login.microsoftonline.com/common/oauth2/v2.0/logout","response_types_supported":["code","id_token","code id_token","id_token token"],"scopes_supported":["openid","profile","email","offline_access"],"issuer":"https://login.microsoftonline.com/{tenantid}/v2.0","claims_supported":["sub","iss","cloud_instance_name","aud","exp","iat","auth_time","acr","nonce","preferred_username","name","tid","ver","at_hash","c_hash","email"],"request_uri_parameter_supported":false,"tenant_region_scope":null,"cloud_instance_name":"microsoftonline.com","cloud_graph_host_name":"graph.windows.net"}
```


xerners:
Hey @iainmcgin and @WilliamDenniss any thoughts on this?

ve7jtb:
Connect requires that the issuer value in the token lead to a .well_known that contains the discovery document with endpoints for a authorization server.  

I know Microsoft has discovery files for individual tenants.

eg
For issuer https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07

wget https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/v2.0/.well-known/openid-configuration

gets you
{"authorization_endpoint":"https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/oauth2/v2.0/authorize","token_endpoint":"https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/oauth2/v2.0/token","token_endpoint_auth_methods_supported":["client_secret_post","private_key_jwt"],"jwks_uri":"https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/discovery/v2.0/keys","response_modes_supported":["query","fragment","form_post"],"subject_types_supported":["pairwise"],"id_token_signing_alg_values_supported":["RS256"],"http_logout_supported":true,"frontchannel_logout_supported":true,"end_session_endpoint":"https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/oauth2/v2.0/logout","response_types_supported":["code","id_token","code id_token","id_token token"],"scopes_supported":["openid","profile","email","offline_access"],"issuer":"https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/v2.0","claims_supported":["sub","iss","cloud_instance_name","aud","exp","iat","auth_time","acr","nonce","preferred_username","name","tid","ver","at_hash","c_hash","email"],"request_uri_parameter_supported":false,"tenant_region_scope":"EU","cloud_instance_name":"microsoftonline.com","cloud_graph_host_name":"graph.windows.net"}

I know that some creative interpretation of the Connect specs has clients going to the common discovery endpoint and using one client ID and effectively getting back a response from a different issuer from the one it thought it made the request to.   This introduces a potentially large number of issues that need to be considered if this were to become a general pattern across IdP.  

is not discovering https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/v2.0 and retreving the keys from https://login.microsoftonline.com/6574d610-008d-4501-83cd-8dfe9dfa0e07/discovery/v2.0/keys then all sorts of bad things can happen.  I suspect that most clients don't do that and that is OK because MS is a good player.

A bad player may take advantage if the client is not checking that 
the issuer it made the request to is the one that it got a reply from. 

The latest Oauth security advice of having per issuer redirect URI is also going to cause Azure trouble if it is not documented.

I note also that {tennantid} is not mentioned in any spec. 

We should talk this over with Mike Jones and perhaps Vittorio.

I am hesitant to make potential security changes without a specification.

On the other hand working with Azure is a goal.

John B.

brandwe:
Hey @WilliamDenniss what are your thoughts on this? I'd like to push our sample to you soon and this will determine if we support MSA and Azure Active Directory multi-tenant scenarios or just single tenant.

brandwe:
Actually, how about this @StevenEWright @WilliamDenniss - what if I add an Issuer Validation flag? Then, for multi-tenant scenarios (where an app can accept more than one tenant) we would tell them to set this issuer validation to off? That is what we do in MSAL today. That way it would validate by default, but we would then have the option for our scenarios?

WilliamDenniss:
Working on this during the IETF hackathon. I've [registered](https://apps.dev.microsoft.com/) my own app client (registration UI worked great btw).  I have 3 comments and a proposal.

### 1/
As you no doubt discovered, the reason AppAuth errors is because the issuer doesn't validate as a URL.  For example, the variable `url` will be `nil` with this ObjC code on iOS:
```
  NSURL *url = [NSURL URLWithString:@"https://login.microsoftonline.com/{tenantid}/v2.0"];
  if (!url)
  {
    NSLog(@"URL invalid");
  }
```

This error is happening during the document validation, rather than later during ID Token validation (though if we got that far, that should fail due to the issuer mismatch due to the tenant substitution extension, more on that topic later).

Issuer is a required field and defined as URL in https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata and https://tools.ietf.org/html/draft-ietf-oauth-discovery-06#section-2. It looks like this discovery doc isn't technically providing a RFC3987 compliant URL in the field "issuer".  I believe if it was percentage-encoded like `"https://login.microsoftonline.com/%7Btenantid%7D/v2.0"` it would be valid.  

**Is it possible to escape the URL in your discovery document?**

### 2/

This is mostly an aside (as I assume you do want discovery to work), but AppAuth can be used without discovery.  In that case, it is possible to specify the OAuth endpoints manually.

Here's what I used:
```
  NSURL *authorizationEndpoint =
      [NSURL URLWithString:@"https://login.microsoftonline.com/common/oauth2/v2.0/authorize"];
  NSURL *tokenEndpoint =
      [NSURL URLWithString:@"https://login.microsoftonline.com/common/oauth2/v2.0/token"];
  OIDServiceConfiguration* configuration =
      [[OIDServiceConfiguration alloc] initWithAuthorizationEndpoint:authorizationEndpoint
                                                       tokenEndpoint:tokenEndpoint];
```

And my example request:
```
  // builds authentication request
  NSArray<NSString *> *scopes = @[OIDScopeOpenID,
                                  OIDScopeProfile,
                                  @"https://outlook.office.com/mail.read"];
  OIDAuthorizationRequest *request =
      [[OIDAuthorizationRequest alloc] initWithConfiguration:configuration
                                                    clientId:clientID
                                                clientSecret:clientSecret
                                                      scopes:scopes
                                                 redirectURL:redirectURI
                                                responseType:OIDResponseTypeCode
                                        additionalParameters:nil];
```

The samples currently require a discovery endpoint – but it wouldn't be too hard to modify them to accept manual configuration. If that actually solved the problem.

### 3/

The late binding tenant OIDC extension perplexes me a little. If I understand correctly, the protocol is that you do discovery on a "common" issuer endpoint, followed by the usual OIDC request. Unusually, the resulting ID Token has an issuer that doesn't exact match that from discovery. Rather, it's a pattern match, with `{tenantid}` in the initially discovered issuer being substituted with the user's tenant.

I gather from this that to validate such an ID Token we'd need a substitution rule for the `https://login.microsoftonline.com` domain to basically allow any value (specifically one matching the regex `[A-Za-z0-9\-]`) to replace `{tenantid}` in the original issuer.  Then, if the developer does discovery against `/common`, it implies they're willing to accept any tenant (in that domain) – but if they do discovery using a tenant-specific issuer it indicates they're only willing to accept that specific tenant. The latter validation logic is standard OIDC, the former is not (and what I'm dubbing the "late binding tenant extension" for lack of a better term).

Is my understanding of the late binding tenant protocol extension correct? Is there a canonical document I should reference?

I'm at IETF so I can ask Mike Jones to clarify these points too (and run any proposed solutions past @ve7jtb).

### Proposal

Ideally I'd like to solve **/3** without an `if (microsoft)` block of code in AppAuth.  One idea on how to solve that would be to implement the late binding tenant behavior as a standard protocol extension.

For example, we could introduce a new discovery document field `issuer_tenant_substitution` which would advertise that tenant substitution was supported by the issuer, and provide the value to be substituted against. For Microsoft, this could be set like so: `"issuer_tenant_substitution": "{tenantid}"`.

Then, AppAuth – and other OIDC libraries that support the late binding tenant extension – can include tenant substitution logic in their ID Token validation code in a generic way.

Aside: is there any plan to propose this extension as an official OIDF standard? Even in draft status, it would be very useful for implementors, particularly around the security considerations to make sure everyone is implementing the validation logic in the same way.

WilliamDenniss:
After conferring with @ve7jtb he added that there may need to be an additional discovery step when validating the id token to get the right keys (as the tenant signing keys apparently can differ to the `/common` signing keys).

Out of the box, AppAuth doesn't validate the signature (trusting the HTTPS connection instead) so that particular issue is moot for this library, but if that changes (or if you're reading this in the context of another library) it may matter.

WilliamDenniss:
Bulk closing old PRs, can re-open if there is activity.

s-limo:
> After conferring with @ve7jtb he added that there may need to be an additional discovery step when validating the id token to get the right keys (as the tenant signing keys apparently can differ to the `/common` signing keys).

The main issue here is not about validating ID token. It's about validating discovery document when iOS tries to parse issuer field:
`NSURL *url = [NSURL URLWithString:@"https://login.microsoftonline.com/{tenantid}/v2.0"];
`
If this issue could be solved (e.g. by configuration), this library could become usable)

PR #114, opened 2017-05-02
-[HTTPServer start:] will no longer fail if it can't bind an IPv6 socket

ksuther:
We had a user with IPv6 completely disabled (on Mac), which was causing `-[HTTPServer start:]` to fail.
Since only the IPv4 loopback address is used, I've made `start:` succeed even if it can't bind an IPv6 socket.

I also added the option of using the IPv6 loopback address (`::1`) in the future if iPv6 should be preferred over IPv4.

### DIFF
--- a/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.h
+++ b/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.h
@@ -72,6 +72,9 @@ typedef enum {
 - (BOOL)start:(NSError **)error;
 - (BOOL)stop;
 
+- (BOOL)hasIPv4Socket;
+- (BOOL)hasIPv6Socket;
+
 // called when a new connection comes in; by default, informs the delegate
 - (void)handleNewConnectionFromAddress:(NSData *)addr
                            inputStream:(NSInputStream *)istr
--- a/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
+++ b/Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
@@ -505,21 +505,17 @@ static void TCPServerAcceptCallBack(CFSocketRef socket, CFSocketCallBackType typ
     addr4.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
     NSData *address4 = [NSData dataWithBytes:&addr4 length:sizeof(addr4)];
 
-    if (kCFSocketSuccess != CFSocketSetAddress(ipv4socket, (CFDataRef)address4)) {
-        if (error) *error = [[NSError alloc] initWithDomain:TCPServerErrorDomain code:kTCPServerCouldNotBindToIPv4Address userInfo:nil];
+    if (kCFSocketSuccess == CFSocketSetAddress(ipv4socket, (CFDataRef)address4)) {
+        if (0 == port) {
+            // now that the binding was successful, we get the port number
+            // -- we will need it for the v6 endpoint and for the NSNetService
+            NSData *addr = (__bridge_transfer NSData *)CFSocketCopyAddress(ipv4socket);
+            memcpy(&addr4, [addr bytes], [addr length]);
+            port = ntohs(addr4.sin_port);
+        }
+    } else {
         if (ipv4socket) CFRelease(ipv4socket);
-        if (ipv6socket) CFRelease(ipv6socket);
         ipv4socket = NULL;
-        ipv6socket = NULL;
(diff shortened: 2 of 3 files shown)


Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
```diff
@@ -585,4 +583,14 @@ - (BOOL)stop {
     return YES;
 }
 
+- (BOOL)hasIPv4Socket
```
StevenEWright: For consistency:

`- (BOOL)hasIPv4Socket {`

Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.m
```diff
@@ -585,4 +583,14 @@ - (BOOL)stop {
...
+{
+    return ipv4socket != nil;
+}
+
+- (BOOL)hasIPv6Socket
```
StevenEWright: For consistency:

`- (BOOL)hasIPv6Socket`

Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.h
```diff
@@ -72,6 +72,9 @@ typedef enum {
 - (BOOL)start:(NSError **)error;
 - (BOOL)stop;
 
+- (BOOL)hasIPv4Socket;
```
StevenEWright: Since `start:` always returns `NO` if it can't create an IPv4 socket, why is this necessary?

StevenEWright: Why not something like:

`@property(nonatomic, readonly) NSURL *preferredLoopbackURL;`

Source/macOS/OIDRedirectHTTPHandler.m
```diff
@@ -76,10 +76,18 @@ - (NSURL *)startHTTPListener:(NSError **)returnError {
       *returnError = error;
     }
     return nil;
-  } else {
+  } else if ([_httpServ hasIPv4Socket]) {
```
StevenEWright: The order of these clauses would cause the library to always use IPv4, since, with your changes, the library always must have an IPv4 socket, but only may have an IPv6 socket, this first test will always pass, leaving no chance for the second test.

ksuther: This was intentional, as I didn't want to change the behavior of the library dramatically without any discussion. Handling a potential IPv6-only case is easy enough to do, I'll add it even though it's unlikely to be hit any time soon.

StevenEWright: That would be very(!) much appreciated.

I do know Apple is concerned with IPv6-only networks ([see here](https://developer.apple.com/support/ipv6/)), so I think the feature may be more valuable than you're giving it credit for. IMHO, the lack of this support is a bug with the existing code.

ksuther:
`start:` will now succeed if IPv6 is available but IPv4 isn't. The callback URL still prefers IPv4 over IPv6 if both are available, but that could be changed by swapping the order of the check in `OIDRedirectHTTPHandler`.

Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.h
ksuther: I don't know that there's much gain to splitting out part of `startHTTPListener:` into a separate property. The method is only a few lines as is.

StevenEWright:
This PR LGTM.

@ksuther Thank you! Were you able to test it, or should I patch it in and try it?

Source/macOS/LoopbackHTTPServer/OIDLoopbackHTTPServer.h
StevenEWright: This comment was more to do with making life easy for consumers of this class - since the only reason consumers care about these values is in the context of determining which loopback address to use.

ksuther:
I tested it by making the IPv4 socket fail, but I didn't actually turn off IPv4 systemwide.

ksuther:
Does this require anything else from me?

WilliamDenniss:
Do you want to squash the commits into one? I can do this for you but then I'm added as the "committer", so thought I'd give you first option.

ksuther:
I squashed them, I hope that's what you were expecting.

WilliamDenniss:
Perfect, thanks.

PR #127, opened 2017-07-10
Set APPLICATION_EXTENSION_API_ONLY to YES on iOS target.

fabiankr:
For many apps it makes sense to use AppAuth (OIDAuthState at least) in app extensions for example when loading new data from a today widget. However, AppAuth doesn’t enable the APPLICATION_EXTENSION_API_ONLY setting.

The only non-safe method call is to +[UIApplication sharedApplication]. With this change AppAuth checks whether the code is run from an app extension or not and then uses objc_msgSend() directly to access the sharedApplication without throwing a compiler error and returns an error when the framework user tried to present the authorization from an extension.

In our app we are using AppAuth for authorizing URL requests in a framework that is shared between our main app targets and app extensions. Without APPLICATION_EXTENSION_API_ONLY set to YES Xcode shows a warning when linking the framework against the app extensions. We don't need to present the authorization view from the extensions, but rely on OIDAuthState.

Using objc_msgSend() might not look especially nice, but there shouldn't be any problems with this. It's at least way better than using -performSelector. Alternatively, NSInvocation could be used, but that would make the code a lot more bloated.


fabiankr:
Selected the wrong branch, sorry.

PR #128, opened 2017-07-10
Don’t set the deployment target at the project level.

fabiankr:
The deployment target should be set at the target level instead.

This raises a warning when adding the framework to projects built against the latest SDKs, because a deployment target of 7.0 is not supported anymore. The deployment target is already set for each target so this setting is redundant and should be removed.


WilliamDenniss:
I'm not seeing this issue, I think it was resolved.

PR #130, opened 2017-07-12
Set APPLICATION_EXTENSION_API_ONLY to YES on iOS target.

fabiankr:
For many apps it makes sense to use AppAuth (OIDAuthState at least) in app extensions for example when loading new data from a today widget. However, AppAuth doesn’t enable the APPLICATION_EXTENSION_API_ONLY setting.

The only non-safe method call is to +[UIApplication sharedApplication]. With this change AppAuth checks whether the code is run from an app extension or not and then uses objc_msgSend() directly to access the sharedApplication without throwing a compiler error and returns an error when the framework user tried to present the authorization from an extension.

In our app we are using AppAuth for authorizing URL requests in a framework that is shared between our main app targets and app extensions. Without APPLICATION_EXTENSION_API_ONLY set to YES Xcode shows a warning when linking the framework against the app extensions. We don't need to present the authorization view from the extensions, but rely on OIDAuthState.

Using objc_msgSend() might not look especially nice, but there shouldn't be any problems with this. It's at least way better than using -performSelector. Alternatively, NSInvocation could be used, but that would make the code a lot more bloated.


WilliamDenniss:
Can you sign the CLA so we can merge? https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md

Can `SFSafariViewController` be used in some categories of app extensions? I'm guessing not in the Today Widget at least.

fabiankr:
Sorry for the delay, I've signed the CLA now.

Actually, OIDAuthorizationUICoordinatorIOS only relies on +sharedApplication when SFSafariViewController doesn't exist (iOS 8 or below) and it uses SFSafariViewController when available. It's up to the consumer of the framework to provide a view controller for the presentation so it can work in the share extension for example.

Once support for iOS 8 is dropped, the code I've added to OIDAuthorizationUICoordinatorIOS can be safely removed while still keeping the APPLICATION_EXTENSION_API_ONLY setting enabled.

julienbodet:
Hi @WilliamDenniss, I'm working on a project using AppAuth and I need this fix! When do you think this PR will be merged? 

Thanks!

WilliamDenniss:
Before this is merged, I would like to see a new example added to exercise this functionality (and so we can add it to the CI tests).

I also wonder if the better approach would be a new framework target with the UICoordinator implementations excluded, designed for extensions. What do you think?

fabiankr:
Adding a new framework target is definitely a cleaner solution than my workaround.
Alternatively, you could move all the base functionality that doesn't rely on UI into a separate framework (like AppAuthCore) and then link that framework from AppAuth. That takes some more effort, but is in the end the better way because each source file is only in one target instead of two.

However, I'm not sure if it's worth the effort. For my use case the workaround was more than good enough.

julienbodet:
Im my case, AppAuth is not imported directly in the app or extension target but in our own framework that we import in our app and extension. If AppAuth define targets for app and extension, I think I will have to do the same in my frameworks (with pod subspec). 

For my use case too, this fix is good enough. But your idea is definitely a cleaner solution, and I like the proposal of @fabiankr.

julienbodet:
@WilliamDenniss I plan to open a new PR picking the work of @fabiankr and resolving conflicts. I'll modify the iOS example by adding an app extension. What's your expectations about this example? An empty Today extension importing AppAuth to demonstrates that the example project compiles, is this sufficient?

WilliamDenniss:
No real reason, just trying to keep the burden low (we need to make sure it
works with CocoaPods, Cathage and hopefully doesn't break anyone who
upgrades).  We now use "canOpenURL" too so the hack surface would increase.

Can you share the results of your Framework tests? If the framework is
viable without too much added hassle it's probably better.

julienbodet:
@WilliamDenniss I just open a PR about the core framework so you can look at it and test it out.

WilliamDenniss:
The underlying issue was resolved by #310 (no need for `objc_msgSend`). Is the APPLICATION_EXTENSION_API_ONLY still needed?

ksuther:
Thanks for the update on this! I believe `APPLICATION_EXTENSION_API_ONLY` is still needed on the new AppAuthCore framework, otherwise at link time it throws a warning:

`ld: warning: linking against a dylib which is not safe for use in application extensions: /path/to/Build/Products/Debug-iphonesimulator/AppAuthCore.framework/AppAuthCore`

If this approach works we could also use an equivalent AppAuthCore_macOS target (I'm happy to do that once we're sure this is working).

WilliamDenniss:
Ok, thanks for checking.

AppAuthCore has none of the platform specific code, so would we need a separate macOS version? I assumed the existing one would work for extensions on all platforms.

WilliamDenniss:
Nice resource, will keep that in mind.

`APPLICATION_EXTENSION_API_ONLY` is now set on the AppAuthCore framework target.

PR #145, opened 2017-09-28
Updates AppAuth for Xcode 9

sergiocampama:
Resolves some project and code warnings.
Updates Swift for iOS tests to Swift 4. This means that iOS tests can only be run with Xcode 9
Tested that Xcode 8.3.3 builds correctly, with the exception of iOS tests (because of Swift 4)


sergiocampama:
Sorry about that, I've now put it back. The ones that are removed are for the tests targets, which are still set to run as Debug, and because Build Active Architecture Only is set to YES in Debug, it still won't build it in fat, so that setting is unnecessary in my opinion (only for tests targets, it's necessary for the framework and static lib)

WilliamDenniss:
LGTM

PR #150, opened 2017-10-05
Pass authorizationRequest additionalParameters to tokenExchangeRequest

jj0b:
A small change so that any additionalParameters on the authorizationRequest would now also get set on the subsequent tokenExchangeRequest. It seems that this is the way AppAuth-Android already functions.


WilliamDenniss:
The token request and the authorization request are different, I'm not sure if everyone would expect that additional authorization request params would be included on the token request. How I'd normally expect a client to do this would be to implement their own token exchange (e.g. by copying the `authStateByPresentingAuthorizationRequest` convenience method and adapting it to their needs).

WilliamDenniss:
The example has a demo for how to do a manual code exchange, and pass additional parameters, it's not a whole lot more work than the automated option.

PR #158, opened 2017-10-11
Resolved the "-Wpartial-availability" warning.

### LINKED ISSUE #146: Does not compile for any target because of unavailable SFSafariViewController
On macOS 10.13.1 Beta
Xcode 9.0 (9A235)
AppAuth 0.9.1
Project does not compile because of  `'SFSafariViewController' is only available on iOS 9.0 or newer`

Changing target from iOS 9 to 10, or 11 does nothing.
Most probably the problem is in Xcode, but anyway, **any hints how to fix this**?
screen shot 2017-09-30 at 12 57 39 pm

WilliamDenniss:
Implemented dismiss functionality for SFAuthenticationSession on iOS 11.
Removed GCC_TREAT_WARNINGS_AS_ERRORS=YES from the Podspec, enabled `-Werror` for TravisCI instead to catch new warnings at the library-developer level.
AppAuth now requires a Base SDK of iOS 11 (still supports iOS 7 as a Deployment Target).

Fixes #146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ xcode_scheme: AppAuth-iOS
 before_script:
   - sudo gem install xcpretty
 script:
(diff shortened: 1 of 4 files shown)


alexzchut:
@WilliamDenniss - Can you please make new tag for this change, because currently is it throwing errors on SFSafariViewController when integrating the framework. Thanks

PR #159, opened 2017-10-11
Corrected documentation errors in OIDTokenResponse.

### LINKED ISSUE #154: Getting the raw token
I'm using AppAuth on iOS, and I need to get the raw token string so that I can use it to make IMAP calls to Gmail.

* `OIDTokenResponse.h` indicates that this is in `accessToken`, but only if the `response_type` includes `token`. However...
* If the `response_type` includes `token`, `OIDAuthorizationRequest` asserts because this type is not supported. **However**...
* Even with a supported response type, `accessToken` is not nil after authentication. ***However***...
* The IMAP server rejects this value.

This is all pretty confusing. It's documented that I can't get a token the way I'm authenticating, but I can't use the documented way because it's unsupported. But I get one anyway, only it doesn't work. Can anyone point me in the right direction?

WilliamDenniss:
– Bunch of params were describing implicit flow behavior for the token response.

Fixes #154


WilliamDenniss:
Fixes #154

PR #163, opened 2017-10-13
Fix compilation on macOS

jseibert:
The macOS deployment target is incorrectly specified, causing the pod to fail to compile.


WilliamDenniss:
Thanks for the PR, this issue is fixed with #170.

PR #166, opened 2017-10-30
Fixes #164 - The iOS Library requires the 9.0 SDK to compile

### LINKED ISSUE #164: Wrong iOS deployment target in podspec
The AppAuth podspec has deployment target set to iOS 7.0 while it requires SafariServices framework which is only available in iOS 9.0 or later. This is causing build failure unless deployment target is bumped manually or via podfile. It would be great to have the deployment target in accordance with used frameworks. See the screenshot for reference:
screen shot 2017-10-25 at 12 03 44 pm

anglinb:
I'm running into this issue when using version `0.90.0`. I have had to manually fix this each time I run pod install.

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -39,7 +39,7 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
   # iOS
   s.ios.source_files      = "Source/iOS/**/*.{h,m}"
-  s.ios.deployment_target = "7.0"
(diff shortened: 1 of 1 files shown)


WilliamDenniss:
Thanks for the PR. This is already fixed in master for iOS (and macOS in #170)  and a new pod version will be released shortly.

anglinb:
Thanks @WilliamDenniss 🙌

Issue #167, opened 2017-10-30
macOS support


eddie129:
In the latest version 0.90.0, @available are added in some methods. But they only handled the iOS part, and caused compile error in macOS。

WilliamDenniss:
Fixed by #170.

PR #168, opened 2017-10-30
#167 Fix macOS support issue

eddie129:
Move methods that use @available(iOS 8.0, *) in OIDURLQueryComponent.m to OIDURLQueryComponent+IOS.m/h, and add corresponded methods for macOS in OIDURLQueryComponent+Mac.m/h.


WilliamDenniss:
Hey, thanks for the PR. The issue was around the availability declarations. Previously this went unnoticed, but the latest version of Xcode is better at validating this.

#170 should fix the issue without needing to duplicate code.

PR #170, opened 2017-10-31
Fixed deployment targets and availability.

WilliamDenniss:
– iOS deployment target is 7.0 for the static lib, and 8.0 for the Framework (the static lib had been set to 8.0 incorrectly).
– macOS deployment target is 10.9 (was on 10.8, but wouldn't have worked).
– Added proper availability flags to OIDURLQueryComponent for macOS.


seungrye:
The readme.md file states that macOS 10.8 is supported. 
The content should also be changed to support 10.9 or later.

PR #174, opened 2017-11-09
Set BITCODE_GENERATION_MODE to 'bitcode'

WilliamDenniss:
- Both Release and Debug were being built with -fembed-bitcode-marker
  which doesn't actually embed bitcode.
- Now -fembed-bitcode is passed which should embed the bitcode correctly.

Should fix https://github.com/google/google-api-objectivec-client-for-rest/issues/171

### DIFF
--- a/AppAuth.xcodeproj/project.pbxproj
+++ b/AppAuth.xcodeproj/project.pbxproj
@@ -1825,6 +1825,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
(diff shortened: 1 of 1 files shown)


WilliamDenniss:
Relevant information:
https://medium.com/@heitorburger/static-libraries-frameworks-and-bitcode-6d8f784478a9
https://stackoverflow.com/a/34965178/72176

I went with the User-Defined Setting rather than the Other C flags option, as the former avoids unused param warnings.

StevenEWright:
Nice! Good to know. Interesting choice on Apple's part that this isn't now the default.

PR #181, opened 2017-11-21
Box fix

VitaliyTeleusov:
1. Fix authorization in box. To get token need to add client_id and client_secret as parameters in POST request. https://developer.box.com/v2.0/reference#token
2. Edit project settings according to AppStore validation.

### DIFF
--- a/AppAuth.xcodeproj/project.pbxproj
+++ b/AppAuth.xcodeproj/project.pbxproj
@@ -1885,6 +1885,7 @@
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
+				SKIP_INSTALL = YES;
 			};
 			name = Debug;
 		};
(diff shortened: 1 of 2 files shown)


WilliamDenniss:
According to [RFC6749](https://tools.ietf.org/html/rfc6749#section-2.3.1) Section 2.3.1, the authorization server MUST support HTTP Basic authentication (which is what AppAuth uses).

```
The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.
```

I'm reluctant to add authorization-server specific hacks to AppAuth when the RFC is very clear on this point.

VitaliyTeleusov:
According to "2.3.1.  Client Password": "Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or othe password-based HTTP authentication schemes)." But " The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI."

WilliamDenniss:
Exactly – this proposed change would add the client credentials to the request-body, which is NOT RECOMMENDED by RFC 6749 Section 2.3.1. AppAuth *is* able to directly utilize HTTP Basic, which is why we follow the recommendations for clients that are able to use HTTP Basic.

Servers that implement RFC 6749 MUST support HTTP Basic for client authentication, per the text I quoted above, so AppAuth needs to only support HTTP Basic.

WilliamDenniss:
Closing this PR, as AppAuth is relying here on a feature that MUST be provided by implementations of RFC 6749, and our policy is not to workaround bugs in the authorization server. I suggest this provider correct their implementation.

PR #138, opened 2017-09-05
Objects for RP-initiated initiated logout

zachmargolis:
These are the basic objects needed to support RP-initiated logout, I'd like to make sure they look ship-shape before adding methods to use them and follow up in another PR

PR #191, opened 2018-01-11
Picking up on @zachmargolis work, and fixing review comments

luksfarris:
From @zachmargolis:

>These are the basic objects needed to support RP-initiated logout, I'd like to make sure they look ship-shape before adding methods to use them and follow up in another PR

I've picked his work, and fixed the comments made by @WilliamDenniss

### DIFF
--- a/Source/AppAuth.h
+++ b/Source/AppAuth.h
@@ -37,6 +37,8 @@
 #import "OIDTokenResponse.h"
 #import "OIDTokenUtilities.h"
 #import "OIDURLSessionProvider.h"
+#import "OIDEndSessionRequest.h"
+#import "OIDEndSessionResponse.h"
 
 #if TARGET_OS_TV
 #elif TARGET_OS_WATCH
--- a/AppAuth.xcodeproj/project.pbxproj
+++ b/AppAuth.xcodeproj/project.pbxproj
@@ -377,6 +377,36 @@
 		60140F801DE4344200DA0DC3 /* OIDRegistrationResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F7F1DE4344200DA0DC3 /* OIDRegistrationResponse.m */; };
 		60140F831DE43BAF00DA0DC3 /* OIDRegistrationRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F821DE43BAF00DA0DC3 /* OIDRegistrationRequestTests.m */; };
 		60140F861DE43CC700DA0DC3 /* OIDRegistrationResponseTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F851DE43CC700DA0DC3 /* OIDRegistrationResponseTests.m */; };
+		A6CEB11A2007E49C009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6CEB11B2007E49D009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6CEB11C2007E49E009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6CEB11D2007E49F009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6CEB11E2007E4A1009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6CEB11F2007E4A2009D492A /* OIDEndSessionRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A6CEB1182007E384009D492A /* OIDEndSessionRequestTests.m */; };
+		A6DEAB832017A7030022AC32 /* OIDEndSessionResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = CF6431F31F228A980075B6B5 /* OIDEndSessionResponse.m */; };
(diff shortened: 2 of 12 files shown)


Source/OIDLogoutResponse.m
```diff
@@ -0,0 +1,84 @@
+/*! @file OIDLogoutResponse.m
+    @brief AppAuth iOS SDK
+    @copyright
+        Copyright 2015 Google Inc. All Rights Reserved.
```
WilliamDenniss: Update all instances `2015 Google Inc. All Rights Reserved.` -> `2017 The AppAuth Authors.`

Source/OIDLogoutResponse.h
```diff
@@ -0,0 +1,62 @@
...
+@property(nonatomic, readonly, nullable) NSString *state;
+
+/*! @internal
+    @brief Unavailable. Please use initWithParameters:.
+ */
+- (instancetype)init NS_UNAVAILABLE;
+
+/*! @brief Designated initializer.
+    @param request The serviced request.
+    @param parameters The decoded parameters returned from the Server.
```
WilliamDenniss: what endpoint is this? rather than "from the Server" we should say "from the XYZ endpoint".

WilliamDenniss:
Thanks for addressing the review comments from #138.  I think this is nearly good to go, but I have a few questions:

1/
Can both you and @zachmargolis please sign the OpenID [CLA](http://openid.net/intellectual-property/contribution-license-agreement/) to cover this contribution?

2/
Do you plan to implement the controllers for logout in addition to the model implemented here?

3/
I don't have much experience with the [OpenID Connect Session Management draft](http://openid.net/specs/openid-connect-session-1_0.html#OPMetadata). How stable is it in general? Do you have a working server implementation to test the client code against? Do you plan to use this client code in production?

zachmargolis:
Re 1/: I signed. Thanks for picking this back up!

Source/OIDLogoutResponse.h
luksfarris: Good point, will rewrite soon!

Source/OIDLogoutResponse.m
luksfarris: Will do, thanks!

luksfarris:
Re 1/: signed as well. Thanks!

2/ I'm doing it right now. Will update the PR in several hours. I'll also rename everything from `Logout` to `EndSession`, to match the spec

3/ We are using okta's oidc, and they, in theory, support the end session flow. We plan to use it in production, as soon as we get it working. This pull request is interesting for us to have it available in the near future

dgommers:
We are also in demand of this feature and I was about to test it with a Keycloak IDC instance. I tried to integrate the fork using Carthage. However, I discovered the newly added files are not attached to all the framework targets. This makes the new functionality unavailable for Carthage users. 

To me it is unclear who should construct an`OIDLogoutRequest` instance and execute it. Could you elaborate on this? Right now it looks quite different from other request/response pairs, for example the `OIDTokenRequest`.

luksfarris:
@dgommers I have updated the pull request adding the targets and fixing the Docs. I am currently renaming it from `Logout` to `EndSession` and adding a new class method called `+performEndSessionRequest:(OIDendSessionRequest *)request completion:(OIDEndSessionCompletion)completion` to class `OIDAuthorizationService`. Do you have a better suggestion of where it should be?

WilliamDenniss:
Thanks for signing the CLA all!

> EDIT2: after further investigation, this looks like it will need a similar protocol to OIDAuthorizationFlowSession, and a class method in OIDAuthState like what we have for authStateByPresentingAuthorizationRequest:presentingViewController:callback:

Yes, you're right.

Previously there was only one place a user-agent (browser / SFAuthenticationSession) was used, now there will be multiple. So whatever way you implement this, the one thing I'm definitely going to be looking for is having the "open URL in user agent / callback" part reused in both cases – I hope there's a good way to generalize that code.

Regarding when we integrate this into master, I think it makes sense to merge this feature into master only when it's usable (models + controller). Do you want a mega PR with the models + controller, OR, we could merge this into a development branch so you can start a new PR with the controller.  A development branch still enables people to test it via Carthage/CocoaPods if they specify the branch so that could be useful.

dgommers:
Got it, thanks. Good to hear that logic can be shared with the `OIDAuthorizationFlowSession`.

Suggestion: A separate protocol for such `resumeAuthorizationFlowWithURL:` flows would help to keep the `AppDelegate` code clean. Implemented by both `OIDAuthorizationFlowSession` and the end session flow class.

Let me know if I can be of any help. I am keeping an eye on this thread.

luksfarris:
I agree with you William, it makes a lot of sense to split this into 2 pull requests. I've renamed the request/response/test classes, and squashed the 3 commits. I'll start working on the other PR right away. 
Thanks for the support!

luksfarris:
I've created issue #195 and PR #196 to address the necessary architecture changes for us to create the RP-Initiated logout controller

WilliamDenniss:
OK, are you ready for me to merge this into a development branch for logout?

luksfarris:
Yes we are!

WilliamDenniss:
I created "logout-dev" can you point the PR at that?

Also can you reword your commit message to describe the actual changes that are going in? Feel free to credit Zach in the body of the message, but I think the first line should focus on the code changes.

#196 can go direct to master, but the rest we can merge into dev-logout until it's complete.

luksfarris:
Done, and done!

Source/AppAuth.h
```diff
@@ -37,6 +37,8 @@
 #import "OIDTokenResponse.h"
 #import "OIDTokenUtilities.h"
 #import "OIDURLSessionProvider.h"
+#import "OIDEndSessionRequest.h"
```
WilliamDenniss: Imports should be in alphabetical order, can you address in a subsequent commit? Here and the framework header.

luksfarris: I will! Thanks William

jaishankar:
@luksfarris @WilliamDenniss Thanks for this logout feature, my understanding is that, after the PR 196 is done, there is going to a  new class method (say. authStateByPresentingEndSessionEndpointRequest) in OIDAuthState which returns the callback with EndSessionEndpointResponse. Is this correct or can you clarify on how the SFAuthenticationSession / browser be invoked for the logout flow?

luksfarris:
@jaishankar with the other pull request you saw, we'd have the ability to create non authentication external user agent requests. This allows us to create a new Logout FlowSession, and have it called in a class method in OIDAuthState. 

How the browser will be invoked, is an open question. SFAuthenticationSession has a terrible user experience, because every time you open it, it asks for a "Sign In" confirmation (therefore our plan of "generic" requests looks and feels weird. With @WilliamDenniss 's implementation of external browsers, we now have other possibilities

PR #193, opened 2018-01-17
Call completion block with redirect mismatch error

grnbeagle:
This PR fixes an iOS 11 issue for supporting redirect URI that redirects.

This is necessary if the client uses a redirect URI that redirects. For example, a provider may not support custom URI so that a backend component needs to redirect from `http://oauth.myapp.com/` to `myapp://`. Because `shouldHandleURL` returns false, we're not able to finish authentication process in iOS 11. This wasn't necessary for iOS 10 as we had to handle the URL manually. 

Let me know if this is appropriate. Other options I considered are: 
* Call openURL instead of resumeAuthFlow ([code](https://github.com/openid/AppAuth-iOS/blob/0.90.0/Source/iOS/OIDAuthorizationUICoordinatorIOS.m#L98)), but this launches a new instance of Safari.
* Pass in a list of accepted redirect URLs - this requires changes in the model code.
* We could rely on iOS to determine whether the given callback URL is valid if/when iOS supports it  down the road: http://www.openradar.me/35796658

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -102,10 +102,6 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (BOOL)resumeAuthorizationFlowWithURL:(NSURL *)URL {
-  // rejects URLs that don't match redirect (these may be completely unrelated to the authorization)
-  if (![self shouldHandleURL:URL]) {
-    return NO;
-  }
   // checks for an invalid state
   if (!_pendingauthorizationFlowCallback) {
     [NSException raise:OIDOAuthExceptionInvalidAuthorizationFlow
@@ -143,6 +139,18 @@ NS_ASSUME_NONNULL_BEGIN
                                   code:OIDErrorCodeOAuthAuthorizationClientError
                               userInfo:userInfo];
       }
+  }
+
+  // rejects URLs that don't match redirect (these may be completely unrelated to the authorization)
+  if (![self shouldHandleURL:URL]) {
+    error = [NSError errorWithDomain:OIDOAuthAuthorizationErrorDomain
+                                code:OIDErrorCodeOAuthInvalidRedirectURI
+                            userInfo:nil];
+    [_UICoordinator dismissAuthorizationAnimated:YES
+                                      completion:^{
+                                        [self didFinishWithResponse:response error:error];
+                                      }];
+    return NO;
   }
 
   [_UICoordinator dismissAuthorizationAnimated:YES


WilliamDenniss:
Hi! Thanks for making this contribution. Can you sign the [OpenID CLA](http://openid.net/intellectual-property/contribution-license-agreement/) so we can consider your contribution?

How were you getting around the URL mis-match before iOS 11? `shouldHandleURL` is still being called right, were you rewriting the URL before passing it to resumeAuthorizationFlowWithURL?

With your change here, when didFinishWithResponse: is [called](https://github.com/openid/AppAuth-iOS/pull/193/files#diff-babcf557cd66620dac5a61fee0ae8baeR151) it may have a non-nil response *and* non-nil error. Are you simply ignoring the error object in your application code? Normally in AppAuth we only ever have one of the objects as non-nil.

WilliamDenniss:
The problem I see with this is that we're potentially exposing a weakness where the client will accept a mis-matched redirect URI.

Another way to solve this would be to use the same web page intermediary to rewrite the outbound request.  e.g.:

AppAuth->Intermediary: OAuth request with `com.example.yourapp`
Intermediary->Authorization Endpoint: 302 redirect, rewriting `com.example.yourapp` to `intermediary.example.com`
Authorization Endpoint->Intermediary: OAuth response to `intermediary.example.com`
Intermediary->AppAuth: 302 redirect to `com.example.yourapp`

This would require zero changes in AppAuth, but the addition of another endpoint on the website to rewrite outbound requests.

Other than the known issues that this change would introduce (i.e. potential for people not to validate the redirect as they should), I'm also worried about issues we don't know about being introduced since this is kind of non-standard (the AS should really be following [RFC 8252](https://tools.ietf.org/html/rfc8252#appendix-A)).  The method I proposed would keep the workaround purely on the side of the intermediary.

grnbeagle:
Thanks for your feedback @WilliamDenniss. Can you clarify your redirect strategy? In the first step, you're suggesting creating OIDAuthorizationRequest instance with a custom URL `com.example.yourapp`, is that right? 

My use case involves being restricted to use a web URL for the request and then be redirected to the native app from the web URL to custom URL.

I understand `shouldHandleURL` addresses the following [spec](https://tools.ietf.org/html/rfc8252#section-8.10): 

>    To protect against a compromised or malicious authorization server
>    attacking another authorization server used by the same app, it is
>    REQUIRED that a unique redirect URI is used for each authorization
>    server used by the app (for example, by varying the path component),
>    and that authorization responses are rejected if the redirect URI
>    they were received on doesn't match the redirect URI in an outgoing
>    authorization request.
> 
>    The native app MUST store the redirect URI used in the authorization
>    request with the authorization session data (i.e., along with "state"
>    and other related data) and MUST verify that the URI on which the
>    authorization response was received exactly matches it.
> 
>    The requirement of Section 8.4, specifically that authorization
>    servers reject requests with URIs that don't match what was
>    registered, is also required to prevent such attacks.

Do you think when AppAuth rejects due to a mismatch, the client needs to be informed about it? Also wouldn't it be helpful to know the decision was not from the authorization server? 

I thought my solution would be acceptable if an error (with the new redirect URL) is returned so that the client can handle accordingly. 

I'm going to think about alternate solutions that may or may not require AppAuth changes. 
- use a universal link as a redirect URI: not sure if Safari opens native app if it's redirected to a universal link. will investigate.  
- use a different mechanism to check if the redirect URL is safe

http://www.openradar.me/35796658 makes me think this is a use case others run into. Let me know if I misunderstood your previous suggestion or have other ideas. Thanks!

WilliamDenniss:
> Do you think when AppAuth rejects due to a mismatch, the client needs to be informed about it? Also wouldn't it be helpful to know the decision was not from the authorization server?

Good question. The issue here is that URIs may be used for many things, not just authorization, that's why AppAuth ignores the URI rather than generating an error. Think of this scenario:

1. User initiates an OAuth flow with AppAuth
2. User is distracted, and then performs some action which also results in a URI being opened in the app *which may be completely fine, though it depends on how the app is designed*.
3. App gets the URI which has nothing to do with AppAuth.
4. Later, user completes the OAuth flow, redirect URI is sent to the app.

We chose to silently ignore this URI in AppAuth which enables step 4 to complete.  An alternative design could be to hard-reject on any invalid incoming error.

More generally speaking, AppAuth may reject many things through it's own logic that are not the decision of the authorization server, that's because we don't necessarily trust any incoming links (e.g. they could be from a malicious authorization server doing some kind of mix-up attack). State mis-match is a good example.

> Thanks for your feedback @WilliamDenniss. Can you clarify your redirect strategy? In the first step, you're suggesting creating OIDAuthorizationRequest instance with a custom URL com.example.yourapp, is that right?

Yes I am, with `com.example.yourapp` as the redirect, and I guess your proxy website as the authorization endpoint. This isn't technically valid RFC 8252, it's a hack and I can't really vouch for its safety, but it might work if you have an intermediary site that's rewriting the request and response in a known way.

> use a universal link as a redirect URI: not sure if Safari opens native app if it's redirected to a universal link. will investigate.

This is actually a far better solution. Assuming you can target iOS 9+ (and iOS 9's distribution is a lot better than it was when we started AppAuth!), this should work perfectly.  Safari will open the native app if it's redirected to a universal link, but I've not actually tested how this works with SFAuthenticationSession, I suspect the app may still be called.

The main point that `To protect against a compromised or malicious authorization server
attacking another authorization server used by the same app, it is
REQUIRED that a unique redirect URI is used for each authorization` is making is that you need to use a unique redirect for each provider. E.g. `com.example.yourapp:/provider1` `com.example.yourapp:/provider2`. This prevents a certain category of mix-up attack (documented [here](http://openid.net/2016/07/16/preventing-mix-up-attacks-with-openid-connect/), section titled "Using the Code Flow to mitigate attacks involving a compromised OP").

WilliamDenniss:
I believe this is out of scope. Can you solve this client side by re-writing the URL before it is given to AppAuth?

Issue #195, opened 2018-01-24
Change the FlowSession and UICoordinator classes to be generic


luksfarris:
As discussed on the PR #191 it will be interesting for future user-agent requests to use the same flowsession and ui coordinator logics. I've created a pull request that allows these to be generic and extended.
This will be necessary for when the RP-Initiated Logout controller is implemented

PR #196, opened 2018-01-24
Making the FlowSessions and UICoordinators generic

luksfarris:
So they can be used for more than just authorization purposes. Relates to issue #195


WilliamDenniss:
This looks good, nice work!

Can we settle on a different naming convention though? I find "UserAgentUI" and "UserAgentFlowSession" slightly odd. How about we go with "external user-agent" names to match the RFC 8252 terminology. So that would mean:

```
OIDExternalUserAgentSession
OIDExternalUserAgentCoordinator
OIDExternalUserAgentCoordinatorMac
dismissExternalUserAgentAnimated:
```

luksfarris:
Good point! I've renamed the classes, methods, and docs to use the external terminology!

WilliamDenniss:
Code itself looks good, but all of the examples are broken with this PR currently. The reason is that they use the `OIDAuthorizationFlowSession` protocol which was removed.

Thinking more broadly than the examples, literally everyone will be impacted by this change. To avoid that, I think we should do the following:

Keep the old `OIDAuthorizationFlowSession` protocol (and keep the new `OIDExternalUserAgentSession`). Existing implementations of `OIDAuthorizationFlowSession` should implement both `OIDAuthorizationFlowSession` and `OIDExternalUserAgentSession` (`resumeAuthorizationFlowWithURL` can simply call `resumeExternalUserAgentFlowWithURL` in the implementation). New flows like logout can implement just `OIDExternalUserAgentSession`.

OIDAuthorizationService.m will need to be updated too like so:

`+ (id<OIDExternalUserAgentFlowSession,OIDAuthorizationFlowSession>)
    presentAuthorizationRequest:(OIDAuthorizationRequest *)request
                  UICoordinator:(id<OIDExternalUserAgentUICoordinator>)UICoordinator
                       callback:(OIDAuthorizationCallback)callback {
`

That way we can avoid most impact of these changes.

luksfarris:
Very good idea @WilliamDenniss . I've just finish implementing it and it works in the iOS example project. I'm making sure it won't break the other ones, and I'll update the PR and rebase it to resolve these conflicts.
Thanks again

Source/OIDAuthorizationService.m
```diff
@@ -171,6 +171,14 @@ - (void)didFinishWithResponse:(nullable OIDAuthorizationResponse *)response
   }
 }
 
+- (void)failAuthorizationFlowWithError:(NSError *)error {
+  [self failAuthorizationFlowWithError:error];
```
dgommers: This method is calling itself right now, causing an infinite loop. Assuming it is there for backwards compatibility with the old protocol it should call the new `[self failExternalUserAgentFlowWithError:error]` instead.

jaishankar: @WilliamDenniss @luksfarris Please let us know if this will be reviewed and merged. Also can you confirm on the invoking user agent flow for RP Initiated logout - https://github.com/openid/AppAuth-iOS/pull/191#issuecomment-367402450

luksfarris: @dgommers I had missed that. Just fixed and rebased
@jaishankar I'll reply your comment on issue #191 soon

luksfarris:
Awesome! Thank you!

PR #206, opened 2018-02-23
Add an accept header to token request

kadikraman:
The specific use case for this is that GitHub OAuth defaults to [returning the token response in xml](https://developer.github.com/apps/building-oauth-apps/authorization-options-for-oauth-apps/), which causes the library to throw a JSON serialization exception. Thankfully, the result will be returned as JSON if the `Accept: application/json` header is set.

AppAuth-Android already has this header for the same reason, as you can see from the [comments](https://github.com/openid/AppAuth-Android/blob/619e9b0f6d27fd393a8332ceb324d8d243082251/library/java/net/openid/appauth/AuthorizationService.java#L503).

This change should not affect any other spec compliant IDPs.

We are maintaining the [react-native-app-auth](https://github.com/FormidableLabs/react-native-app-auth) library which bridges this and the Android libraries to React Native, hence our desire to expose an api with a consistent behaviour across both platforms 😊

### DIFF
--- a/Source/OIDTokenRequest.m
+++ b/Source/OIDTokenRequest.m
@@ -264,6 +264,8 @@ static NSString *const kAdditionalParametersKey = @"additionalParameters";
   static NSString *const kHTTPContentTypeHeaderKey = @"Content-Type";
   static NSString *const kHTTPContentTypeHeaderValue =
       @"application/x-www-form-urlencoded; charset=UTF-8";
+  static NSString *const kHTTPAcceptHeaderKey = @"Accept";
+  static NSString *const kHTTPAcceptHeaderValue = @"application/json";
 
   NSURL *tokenRequestURL = [self tokenRequestURL];
   NSMutableURLRequest *URLRequest = [[NSURLRequest requestWithURL:tokenRequestURL] mutableCopy];
@@ -272,6 +274,13 @@ static NSString *const kAdditionalParametersKey = @"additionalParameters";
 
   OIDURLQueryComponent *bodyParameters = [self tokenRequestBody];
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
+
+  /**
+   * GitHub will only return a spec-compliant response if JSON is explicitly defined
+   * as an acceptable response type. This is essentially harmless for all other
+   * spec-compliant IDPs.
+   */
+  [httpHeaders setObject:kHTTPAcceptHeaderValue forKey:kHTTPAcceptHeaderKey];
 
   if (_clientSecret) {
     NSString *credentials = [NSString stringWithFormat:@"%@:%@", _clientID, _clientSecret];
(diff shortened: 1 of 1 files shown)


Source/OIDTokenRequest.m
```diff
@@ -273,6 +275,13 @@ - (NSURLRequest *)URLRequest {
   OIDURLQueryComponent *bodyParameters = [self tokenRequestBody];
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
 
+  /**
+   * GitHub will only return a spec-compliant response if JSON is explicitly defined
```
WilliamDenniss: As a rule, AppAuth doesn't support special casing for non-spec compliant implementations of RFC 6749. We need to take a strong stance on this, to avoid bloating the library – and also, there is no reason for people *not* to support RFC 6749 correctly.

That said, since the response should be JSON, I don't mind sending this header. Please change the comment so that the justification isn't something that goes against our policies :-)

Source/OIDTokenRequest.m
```diff
@@ -273,6 +275,13 @@ - (NSURLRequest *)URLRequest {
   OIDURLQueryComponent *bodyParameters = [self tokenRequestBody];
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
 
+  /**
```
WilliamDenniss: Please use `//` style comments here.

kadikraman:
@WilliamDenniss thanks for your review. I've updated the comment and signed the CLA electronically.

kadikraman:
@WilliamDenniss any plans to get this merged?

StevenEWright:
William is not able to review this week. I will ping him about this next week!

WilliamDenniss:
On further consideration, I believe this is out of scope for AppAuth. Our policy is not to accept vendor-specific changes, and none have been accepted to date.

When I read RFC2616 Section 14.1 which defines the Accept behavior, the purpose of this header is for endpoints that can offer multiple formats. However, the OAuth specification does not allow for multiple formats, JSON is the _one and only_ format specified, therefore it is an irrelevant header as far as OAuth is concerned.  If we start sending Accept headers, do we also need to handle a `406 Not Acceptable` response, for example?

While the code looks innocent, it is a slippery slope. If we accept this (no pun intended), then should we accept other vendor-specific hacks?

@StevenEWright had an idea that we could allow user-settable headers in a similar way that we allow for additional parameters, which is probably the correct way forward.  Enabling extensibility of the AppAuth library *is* in-scope, and such a change could be used to support a broad range of uses-cases where additional headers are needed, that are outside the scope of the protocols we directly support.

Until that time, for users who need this, I would recommend using a vendor-specific fork of AppAuth that includes this workaround.

kadikraman:
Fair enough. The annoying this is that [AppAuth-Android](https://github.com/openid/AppAuth-Android/blob/619e9b0f6d27fd393a8332ceb324d8d243082251/library/java/net/openid/appauth/AuthorizationService.java#L503) _does_ support this, which makes the bug incredibly frustrating for React Native. Even though GitHub is clearly in the wrong in this case, I think I have an ever lesser chance of getting them to fix their API, so I guess that's that 😄

The user-settable headers idea sounds sensible, hope it goes forward.

meherranjan:
Just curious if user-settable headers are available in the latest release?

PR #213, opened 2018-03-08
Remove the redirect_uri parameter from token refresh

### LINKED ISSUE #207: Gmail error: Parameter not allowed for this message type: redirect_uri
I am using AppAuth (v0.92.0) on iOS and macOS. The app has already performed OAuth 2 authorization and received tokens from Gmail.

Recently, I have started seeing the error shown below intermittently on both iOS and macOS, when attempting to call `performActionWithFreshTokens:`.

    Parameter not allowed for this message type: redirect_uri

The console log output has the following message:

    OIDAuthState:updateWithTokenResponse should not be called in an error state [Error Domain=org.openid.appauth.oauth_token Code=-2 "invalid_request: Parameter not allowed for this message type: redirect_uri" UserInfo={OIDOAuthErrorResponseErrorKey={
    error = "invalid_request";
    "error_description" = "Parameter not allowed for this message type: redirect_uri";
    }, NSLocalizedDescription=invalid_request: Parameter not allowed for this message type: redirect_uri, NSUnderlyingError=0x1c8050920 {Error Domain=org.openid.appauth.remote-http Code=400 "{
      "error" : "invalid_request",
      "error_description" : "Parameter not allowed for this message type: redirect_uri"
    }" UserInfo={NSLocalizedDescription={
      "error" : "invalid_request",
      "error_description" : "Parameter not allowed for this message type: redirect_uri"
    }}}}] callupdateWithAuthorizationResponse with the result of the fresh authorization responsefirst

This error goes away after some time.

**Question:** Is there anything else  I would need to check before making API calls with IMAP server using OAuth? The examples seem to indicate that only `performActionWithFreshTokens:` is required for ensuring fresh tokens before making API calls, unless I missed something.

Any help would be greatly appreciated!

WilliamDenniss:
On close inspection of [RFC 6749 Section 6](https://tools.ietf.org/html/rfc6749#section-6), the `redirect_uri` isn't required and may cause issues on servers that are not expecting it.

Fixes #207

### DIFF
--- a/Source/OIDAuthState.m
+++ b/Source/OIDAuthState.m
@@ -411,7 +411,7 @@ static const NSUInteger kExpiryTimeTolerance = 60;
       initWithConfiguration:_lastAuthorizationResponse.request.configuration
                   grantType:OIDGrantTypeRefreshToken
           authorizationCode:nil
-                redirectURL:_lastAuthorizationResponse.request.redirectURL
+                redirectURL:nil
                    clientID:_lastAuthorizationResponse.request.clientID
                clientSecret:_lastAuthorizationResponse.request.clientSecret
                       scope:_lastAuthorizationResponse.request.scope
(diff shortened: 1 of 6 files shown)


jainricha:
This fix is giving errors: Null passed to a callee that requires a non-null

WilliamDenniss:
Thanks @jainricha, that's fixed now.

PR #218, opened 2018-03-09
Adding core framework

julienbodet:
As discussed in #130 here is what I did to add a core framework. I'm not sure about all the build settings, I just left the defaults when I created the targets. I tested to install `AppAuth/Core` with Cocoapods in the app extension I added to the example and the build succeeds. I also added some UI to the widget to retrieve the user info with the authState saved in User Defaults. Just one thing I don't understand is that I'm able to import `AppAuth.h` in the extension. I should only see AppAuthCore.h. Maybe related to the podspec settings...


WilliamDenniss:
I think there is a simpler way to do this that doesn't require moving any files around, and that is just moving the platform specific files into a sub-spec.

I want to see a solution that allows users who need this to separate out the files they don't need, without increasing the complexity of AppAuth for the rest. I don't think we really want a hierarchy of frameworks that people need to understand (e.g. AppAuth.h that includes AppAuthCore.h).

Something like this:
```
  # no change
  s.source_files = "Source/*.{h,m}"

  spec.subspec 'Core' do |core|
     # subspec for users who don't want the authentication code 
  end

  spec.subspec 'Auth' do |auth|
    # iOS	
    auth.ios.source_files      = "Source/iOS/**/*.{h,m}"	
    auth.ios.deployment_target = "7.0"	  s.ios.deployment_target = "7.0"
    auth.ios.framework         = "SafariServices"	
    # macOS	
    auth.osx.source_files = "Source/macOS/**/*.{h,m}"	
    auth.osx.deployment_target = '10.9'
  end
```

Based on this design here: http://www.dbotha.com/2014/12/04/optional-cocoapod-dependencies/, note the "Lite" subspec which is equivalent to our "Core"

In our case we wouldn't set `default_subspec`, as we want the default to include everything.

WilliamDenniss:
Thanks for the work! Closing this in favor of #318 which implements my suggestion above to use subspecs.

Issue #198, opened 2018-01-30
OIDAuthorizationUICoordinatorIOS not compiling for App Extensions


julienbodet:
I imported AppAuth via Cocoapods in my swift project and the library is not compiling because of the call to `[UIApplication sharedApplication]` in `presentAuthorizeRequest:session:` and leads to the following error:

![error](https://user-images.githubusercontent.com/3465275/35542101-34ada300-052c-11e8-94f2-95b3ce317a74.png)

It would be great to add a mechanism to take into account if we are building for an app extension or not.

EDIT: I see that the PR #130 propose a solution. When will this PR be merged?

WilliamDenniss:
For now I would recommend forking AppAuth and merging that PR if you need this.

I think the correct way to support extensions would be to have a separate AppAuth framework target designed specifically for extensions.

schroepf:
@WilliamDenniss I ran into the same issue and noticed that for my use case the extension didn't need any of the "platform" specific code in the "ios" and "macos" subdirectories, so I created a new target and built a framework (called AppAuthCore) from only the code in the "Source" directory and none of the subdirecotries (macOS or iOS). So that now I can use the AppAuth framework in the iOS app and the AppAuthCore framework for the extension... Now I wonder if this solution has any chance of making it into the official AppAuth-iOS code and if it is worth the effort to create a pull request?

schroepf:
Oh... I just noticed that @julienbodet did basically the exact same thing in #218...

julienbodet:
Hi @schroepf, yes I created a PR a while ago, but it has now a lot of conflicts. Maybe it's easier to create a new pull request. @WilliamDenniss It would be very nice to have that fixed in an upcoming release, is there any chance we could include that in the 1.0.0? For my use case, it is the only bug that prevents me to use the official AppAuth and force me to maintain a fork. Thanks!

WilliamDenniss:
This is a very valid use-case. The reason I stalled on #218 is that it changed a lot, and impacted people who don't need the separation, which I'd like to avoid.

I did some research into subspecs, and added a review comment there in the PR. Perhaps we can address that concern fairly easily, and deliver this functionality.

schroepf:
Oh, didn't know that something like "subspecs" exist in Cocoapods, that's nice! I created PR #310 which creates a "Core" and an "Auth" subspec (as @WilliamDenniss suggested in @julienbodet's pull request #218 ).

[...]

PR #224, opened 2018-03-18
Fix app extension build error

julienbodet:
This PR fixes issue #198. I continued the work of the pull request #130 and added extensions to the examples.


Source/UIApplication+openURL.h
```diff
@@ -0,0 +1,35 @@
+/*! @file UIApplication+openURL.m
```
StevenEWright: I wonder if it would be better to split these four substantive changes out into a separate PR?

julienbodet: So you want a PR for adding the extensions to examples and another one which fix the compile error?

StevenEWright: Well, I think this is a valuable change, and that we should definitely address the use case. My concern is that the size of it may prevent a timely review. The “real” change here is only the UIApplication+openURL category, and the two files that make use of it. I totally agree that we should add examples - but I’m not sure we need to deal with that in the same change. Is splitting it something you’d be willing to do?

@WilliamDenniss WDYT?

jwells89:
Any updates on this PR? We're using AppAuth in a Today extension and hope to see this get merged soon.

WilliamDenniss:
Resolved by #320 thank you for all the work on this!

PR #228, opened 2018-04-09
Azure AD sends expires_in as string instead of number

michallaskowski:
Hi,
I noticed I was getting errors from my apps after some time. Then I noted that token expiration dates were not set in my `OIDAuthState` and I had a look with Charles Proxy what is going on:
Microsoft is sending responses where `expires_in` is a string containing number of seconds. Because this was a string, it was not decoded properly. It will for sure be easier to add support in this library, than to make Microsoft change their API. (They are changing it anyway to v2, but it does not support yet what we need inside the company, so we need to stick with Azure AD)

So - simple `NSNumberFormatter` here was used.

Link to MS docs showing that they are returning `expires_in` as string: https://docs.microsoft.com/pl-pl/azure/active-directory/develop/active-directory-protocols-oauth-code#successful-response-1

### DIFF
--- a/Source/OIDFieldMapping.m
+++ b/Source/OIDFieldMapping.m
@@ -115,11 +115,20 @@
 
 + (OIDFieldMappingConversionFunction)dateSinceNowConversion {
   return ^id _Nullable(NSObject *_Nullable value) {
-    if (![value isKindOfClass:[NSNumber class]]) {
+    long long timeInterval = 0;
+    BOOL valueMapped = NO;
+    if ([value isKindOfClass:[NSNumber class]]) {
+      timeInterval = [(NSNumber *)value longLongValue];
+      valueMapped = YES;
+    } else if ([value isKindOfClass:[NSString class]]) {
+      timeInterval = [(NSString *)value longLongValue];
+      valueMapped = YES;
+    }
+    if (!valueMapped) {
       return value;
     }
-    NSNumber *valueAsNumber = (NSNumber *)value;
-    return [NSDate dateWithTimeIntervalSinceNow:[valueAsNumber longLongValue]];
+
+    return [NSDate dateWithTimeIntervalSinceNow:timeInterval];
   };
 }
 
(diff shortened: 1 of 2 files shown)


brockallen:
Did you report this to Microsoft? They really should be following spec.

iainmcgin:
@selfissued could you route this to the appropriate team at Microsoft? As per [Section 5.1](https://tools.ietf.org/html/rfc6749#section-5.1) of the OAuth2 spec, `expires_in` should be encoded into JSON as a number, not a string:

> Parameter names and string values are included as JSON strings. Numerical values are **included as JSON numbers**.

iainmcgin:
@WilliamDenniss I'll let you decide whether you want to merge this or not. It looks like AppAuth-Android is already tolerant of this (code [here](https://github.com/openid/AppAuth-Android/blob/master/library/java/net/openid/appauth/TokenResponse.java#L222)), as the JSON utility we use will [attempt to coerce strings to numbers](https://android.googlesource.com/platform/dalvik/+/f2f7880a40aed1c8d9c27db49226e47396556aac/libcore/json/src/main/java/org/json/JSON.java#68).

michallaskowski:
@brockallen FYI, Microsoft already created a second version of the api, where they follow the specs. It is just not ready for use for some cases (https://docs.microsoft.com/pl-pl/azure/active-directory/develop/active-directory-v2-limitations), and I can't wait for it to be ready or for them to provide a fix to an older api. Worst case scenario I will use my fork as a dependency for the project I work on. Anyway, I will create an issue with them about the problem, tomorrow if I find time.

WilliamDenniss:
Our policy is generally not to accept vendor-specific workarounds when the spec is clear, which it is in this case.

```
A.14.  "expires_in" Syntax

   The "expires_in" element is defined in Sections 4.2.2 and 5.1:

     expires-in = 1*DIGIT
```

Is there a strong argument to "Be liberal in what you accept, and conservative in what you send" here though do you think?

WilliamDenniss:
Can you sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/)?

selfissued:
I've brought this bug to the attention of the team.  Thanks for letting me know about this.

WilliamDenniss:
Closing this on the assumption that the provider issue was resolved, please re-open if needed.

PR #238, opened 2018-05-08
Show id_token in log

salmonz:
I noticed when I was following the example, I was watching for the id_token to appear in the log, as this is what is most important in the OIDC protocol, but only saw the access token. This patch shows the id_token for both manual code exchange and auto.

### DIFF
--- a/Source/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/iOS/OIDExternalUserAgentIOS.m
@@ -109,12 +109,12 @@ static id<OIDSafariViewControllerFactory> __nullable gSafariViewControllerFactor
     _authenticationVC = authenticationVC;
     openedSafari = [authenticationVC start];
   } else if (@available(iOS 9.0, *)) {
-      SFSafariViewController *safariVC =
-          [[[self class] safariViewControllerFactory] safariViewControllerWithURL:requestURL];
-      safariVC.delegate = self;
-      _safariVC = safariVC;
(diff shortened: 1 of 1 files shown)


WilliamDenniss:
Can you sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/)?

salmonz:
Signed

Examples/Example-iOS_Swift-Carthage/Source/AppAuthExampleViewController.swift
```diff
@@ -396,6 +397,7 @@ extension AppAuthExampleViewController {
             if let authState = authState {
                 self.setAuthState(authState)
                 self.logMessage("Got authorization tokens. Access token: \(authState.lastTokenResponse?.accessToken ?? "DEFAULT_TOKEN")")
+                self.logMessage("Got id_token token. id_token: \(authState.lastTokenResponse?.idToken ?? "DEFAULT_TOKEN")")
```
WilliamDenniss: Can we handle the nil case better? Seems factually incorrect to state that we got an id_token when we didn't. I don't see how printing "DEFAULT_TOKEN" helps anyone. 

Same comment could apply to the accessToken, although I would normally expect it to be nonnull.

WilliamDenniss:
Bulk closing old PRs, can re-open if there is activity.

PR #248, opened 2018-06-07
Add support for 4xx errors

### LINKED ISSUE #234: invalid_client error may return a 401 on token request and then return OIDOAuthTokenErrorDomain instead of OIDErrorCodeServerError
Referring to [RFC 6749 section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2):

> The authorization server responds with an HTTP 400 (Bad Request) status code **(unless specified otherwise)**

For an `invalid_client` error it is specified that the server may return a 401, but actually AppAuth only generates an [OIDOAuthTokenErrorDomain error on 400 and 429 status code](https://github.com/openid/AppAuth-iOS/blob/master/Source/OIDAuthorizationService.m#L289-L325), OIDErrorCodeServerError otherwise.

So an `invalid_client` error with 401 will be considered as an `OIDErrorCodeServerError` error, but  I think it should be an OAuth error (`OIDOAuthTokenErrorDomain`).

WilliamDenniss:
401 is expected per RFC 6749 Section 5.2 for the invalid_client code.
Extensions to OAuth may use other error codes (presumably still in the 4xx range).

Fixes #234

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -301,16 +301,14 @@ NS_ASSUME_NONNULL_BEGIN
       NSError *serverError =
           [OIDErrorUtilities HTTPErrorWithHTTPResponse:HTTPURLResponse data:data];
 
-      // HTTP 400 may indicate an RFC6749 Section 5.2 error response.
-      // HTTP 429 may occur during polling for device-flow requests for the slow_down error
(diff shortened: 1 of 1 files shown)


StevenEWright:
LGTM, presuming you are certain of the discussion on the issue with @julienbodet

PR #250, opened 2018-06-11
Redact auth tokens in description methods

### LINKED ISSUE #77: Truncate auth tokens in description methods
We should truncate the auth tokens in the `description` methods.

This can help avoid inadvertent logging of sensitive tokens. Developers still have access to the raw tokens through the properties provided for that purpose should they need.

CraigLn:
This PR fixes resolves #77


AppAuth.xcodeproj/project.pbxproj
```diff
@@ -391,6 +391,9 @@
 		60140F801DE4344200DA0DC3 /* OIDRegistrationResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F7F1DE4344200DA0DC3 /* OIDRegistrationResponse.m */; };
 		60140F831DE43BAF00DA0DC3 /* OIDRegistrationRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F821DE43BAF00DA0DC3 /* OIDRegistrationRequestTests.m */; };
 		60140F861DE43CC700DA0DC3 /* OIDRegistrationResponseTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 60140F851DE43CC700DA0DC3 /* OIDRegistrationResponseTests.m */; };
+		A5EEF20220CF0E850044F470 /* OIDTokenUtilitiesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A5EEF1FD20CF07760044F470 /* OIDTokenUtilitiesTests.m */; };
```
StevenEWright: It looks like this file may have been added to the wrong target accidentally?

I think this is also why you had to #pragma GCC diagnostic ignored "-Wgnu" - but the other tests don't.

Could you please remove the test from the library's sources?

Source/OIDTokenUtilities.m
```diff
@@ -48,4 +48,8 @@ + (NSData *)sha265:(NSString *)inputString {
   return sha256Verifier;
 }
 
++ (NSString *)truncate:(NSString *)inputString {
+    return [inputString length] >= 6 ? [[inputString substringToIndex:6] stringByAppendingString:@"..."] : @"...";
```
StevenEWright: Bit of a nit here - but the current case of length == 6 is a bit degenerate... we probably shouldn't show the ellipsis when we aren't actually removing any characters. Also, maybe instead of the ellipsis for < the threshold we could just say "[redacted]" to be clear?

How about this.. ?

`return [inputString length] >= 9 ? [[inputString substringToIndex:6] stringByAppendingString:@"..."] : @"[redacted]";`

UnitTests/OIDTokenUtilitiesTests.m
```diff
@@ -0,0 +1,46 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <XCTest/XCTest.h>
+
+#import "Source/OIDTokenUtilities.h"
+
+// Ignore warnings about "Use of GNU statement expression extension" which is raised by our use of
```
StevenEWright: I think if you remove this from the library target's sources you won't need this, as discussed in the comment for project.pbxproj

Source/OIDTokenUtilities.m
```diff
@@ -49,7 +49,7 @@ + (NSData *)sha265:(NSString *)inputString {
 }
 
 + (NSString *)truncate:(NSString *)inputString {
-    return [inputString length] >= 6 ? [[inputString substringToIndex:6] stringByAppendingString:@"..."] : @"...";
+    return [inputString length] >= 9 ? [[inputString substringToIndex:6] stringByAppendingString:@"..."] : @"[redacted]";
```
StevenEWright: Sorry for the nits, but a few minor code formatting issues...

- We always use dot notation for properties, so `inputString.length` instead of `[inputString length]`.
- We only use 2 spaces for indentation, not 4 (please fix here and in the tests).
- We have a max line length of 100 columns. Line continuations are indented +4, so this should be:

```
+ (NSString *)truncate:(NSString *)inputString {
  return [inputString length] >= 9
      ? [[inputString substringToIndex:6] stringByAppendingString:@"..."] : @"[redacted]";
}

```

StevenEWright:
LGTM!

Source/OIDTokenUtilities.m
```diff
@@ -48,4 +48,9 @@ + (NSData *)sha265:(NSString *)inputString {
   return sha256Verifier;
 }
 
++ (NSString *)truncate:(NSString *)inputString {
+  return inputString.length >= 9
```
WilliamDenniss: Can you convert this non-trivial multi-line ternary statement into a regular if-statement?

WilliamDenniss:
I also wonder if people will realize that "..." means redacted, and not just part of the token.

Would:
```
- (void)testTruncate {
  XCTAssertEqualObjects([OIDTokenUtilities truncate:@"0123456789"], @"012345…[redacted]", @"");
}
```

Be more clear?

CraigLn:
Good point on the consistency of returned values.  Updated those, and renamed.

StevenEWright:
🍻 +1 ... Thanks Craig!!

PR #253, opened 2018-06-15
Create issue templates for bugs and feature requests.

StevenEWright:
The only change to the defaults is a question about iOS 12 (since we've gotten 4 or so issue requests about it by now).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -5,7 +5,7 @@ about: Create a report to help us improve
 ---
 
 **Are you filing an issue about iOS 12?**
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Are you filing an issue about iOS 12?**
+Please see preliminary support for iOS 12 from PR: https://github.com/openid/AppAuth-iOS/pull/246 before filing iOS 12 issues. Thanks.
```
WilliamDenniss: I'd use https://github.com/openid/AppAuth-iOS/tree/ios12beta as the link, people can point their preferred dependency manager at that.

StevenEWright: Done. PTAL?

PR #262, opened 2018-06-29
Add a description to errors, even if they have an underlyingError with a description

### LINKED ISSUE #261: Ensure all NSError objects have a description field
In AppAuth currently, many errors have an `underlyingError` property set with highly useful information, but a nil description. This means that `.localizedDescription` doesn't return useful information (though `.description` does).

This currently makes AppAuth harder to debug, as the error message is not always apparent, and may make it appear to developers that the problem is in AppAuth rather than their own code or configuration (for example, #256).

Add a description to every NSError object, either set to the underlyingError's `localizedDescription`, or some other useful string.

WilliamDenniss:
Fixes #261

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -88,10 +88,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)cancel {
   [_externalUserAgent dismissExternalUserAgentAnimated:YES completion:^{
-      NSError *error = [OIDErrorUtilities
-                        errorWithCode:OIDErrorCodeUserCanceledAuthorizationFlow
-                        underlyingError:nil
-                        description:nil];
+      NSError *error = [OIDErrorUtilities errorWithCode:OIDErrorCodeUserCanceledAuthorizationFlow
+                                        underlyingError:nil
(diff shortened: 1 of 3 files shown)


Source/OIDAuthorizationService.m
```diff
@@ -558,9 +585,14 @@ + (void)performRegistrationRequest:(OIDRegistrationRequest *)request
       }
 
       // not an OAuth error, just a generic server error
+      NSString *errorDescription =
+          [NSString stringWithFormat:@"Non-200/201 HTTP response (%d) making registration request "
+                                     "to '%@'.",
+                                     (int)HTTPURLResponse.statusCode,
```
StevenEWright: The only reason I'm mentioning this is because I seem to recall there being an optional warning about loss of precision here on macOS, iOS, tvOS since [NSInteger is a long on those platforms](https://developer.apple.com/documentation/objectivec/nsinteger?language=objc).

WilliamDenniss: There was a warning about loss of precision, which is why I added the cast here. I considered casting to (long) instead, but HTTP status codes well fit within `int`, so it seemed fine to go that way.

Issue #233, opened 2018-05-02
client_secret_basic header value invalid when special characters are used


panva:
According to https://tools.ietf.org/html/rfc6749#section-2.3.1 (+appendix) the client_id and client_secret values are to be `application/x-www-form-urlencoded`-encoded before being used the basic authentication scheme as username and password. This is [currently not the case](https://github.com/openid/AppAuth-iOS/blob/0.92.0/Source/OIDTokenRequest.m#L277-L282).

Given a client_id value of `1PpG/Q 1` and client_secret value of `z/tZ9VwFZqApmIQ+ZH1I5pLk/uB4ud:X2/8bL+wfFTt1rFw=` the Authorization header should be processed as follows

```
// Authorization: Basic base64(formEncode(1PpG/Q 1):formEncode(z/tZ9VwFZqApmIQ+ZH1I5pLk/uB4ud:X2/8bL+wfFTt1rFw=))

// Authorization: Basic base64(1PpG%2FQ+1:z%2FtZ9VwFZqApmIQ%2BZH1I5pLk%2FuB4ud%3AX2%2F8bL%2BwfFTt1rFw%3D)

Authorization: Basic MVBwRyUyRlErMTp6JTJGdFo5VndGWnFBcG1JUSUyQlpIMUk1cExrJTJGdUI0dWQlM0FYMiUyRjhiTCUyQndmRlR0MXJGdyUzRA==
```

This is very likely to cause issues with clients using this scheme with a provider that also omits this requirement and has identifier or a secret with special characters.

WilliamDenniss:
Agree this is a bug.

Deep links:
OAuth2-flavored form-urlencoded: https://tools.ietf.org/html/rfc6749#appendix-B
Referenced spec: https://www.w3.org/TR/1999/REC-html401-19991224/interact/forms.html#h-17.13.4.1

We can do something like: https://github.com/openid/AppAuth-iOS/blob/d27db29ce26cf1d1c390034e94b00d44af3da750/Source/OIDURLQueryComponent.m#L132 just need to specify the right set of reserved characters.

My advise would be to percent encode space, rather than as "+" as it's a valid encoding, and less ambiguous (and we don't need the "+"-encoding for aesthetic reasons).

panva:
> My advise would be to percent encode space

As seen by the example in appendix b, that's the way to go.

PR #266, opened 2018-07-08
Percent encode special characters from client ID and client secret in…

julienbodet:
… Basic Authorization header

This pull request resolves issue #233.

I added code to percent encode all characters other than alphanumeric ones included in `URLQueryAllowedCharacterSet`. I also followed @WilliamDenniss's recommendation to percent encode space rather than replacing it by "+".

### DIFF
--- a/Source/OIDTokenRequest.m
+++ b/Source/OIDTokenRequest.m
@@ -286,9 +286,9 @@ static NSString *const kAdditionalParametersKey = @"additionalParameters";
     
     // Percent encode all non-alphanumeric characters.
     NSString *encodedClientID =
-        [_clientID stringByAddingPercentEncodingWithAllowedCharacters: allowedCharacters];
+        [_clientID stringByAddingPercentEncodingWithAllowedCharacters:allowedCharacters];
     NSString *encodedClientSecret =
-        [_clientSecret stringByAddingPercentEncodingWithAllowedCharacters: allowedCharacters];
+        [_clientSecret stringByAddingPercentEncodingWithAllowedCharacters:allowedCharacters];
 
     NSString *credentials =
         [NSString stringWithFormat:@"%@:%@", encodedClientID, encodedClientSecret];
--- a/Source/OIDTokenRequest.m
+++ b/Source/OIDTokenRequest.m
@@ -22,6 +22,7 @@
 #import "OIDScopeUtilities.h"
 #import "OIDServiceConfiguration.h"
 #import "OIDURLQueryComponent.h"
+#import "OIDTokenUtilities.h"
 
 /*! @brief The key for the @c configuration property for @c NSSecureCoding
  */
@@ -274,22 +275,10 @@ static NSString *const kAdditionalParametersKey = @"additionalParameters";
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
 
   if (_clientSecret) {
-    // https://tools.ietf.org/html/rfc6749#appendix-B
-    // Note: Space character is percent encoded, rather than as "+" as it's a valid encoding,
-    // and less ambiguous (and we don't need the "+"-encoding for aesthetic reasons)
+    // https://tools.ietf.org/html/rfc6749#section-2.3.1
+    NSString *encodedClientID = [OIDTokenUtilities formUrlEncode:_clientID];
+    NSString *encodedClientSecret = [OIDTokenUtilities formUrlEncode:_clientSecret];
     
-    // Starts with the standard URL-allowed character set.
-    NSMutableCharacterSet *allowedCharacters =
-        [[NSCharacterSet URLQueryAllowedCharacterSet] mutableCopy];
-    // Keep only alphanumeric characters.
-    [allowedCharacters formIntersectionWithCharacterSet:[NSCharacterSet alphanumericCharacterSet]];
-    
-    // Percent encode all non-alphanumeric characters.
-    NSString *encodedClientID =
-        [_clientID stringByAddingPercentEncodingWithAllowedCharacters:allowedCharacters];
-    NSString *encodedClientSecret =
-        [_clientSecret stringByAddingPercentEncodingWithAllowedCharacters:allowedCharacters];
-
     NSString *credentials =
         [NSString stringWithFormat:@"%@:%@", encodedClientID, encodedClientSecret];
     NSData *plainData = [credentials dataUsingEncoding:NSUTF8StringEncoding];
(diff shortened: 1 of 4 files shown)


Source/OIDTokenRequest.m
```diff
@@ -274,7 +274,24 @@ - (NSURLRequest *)URLRequest {
...
-    NSString *credentials = [NSString stringWithFormat:@"%@:%@", _clientID, _clientSecret];
+    // https://tools.ietf.org/html/rfc6749#appendix-B
+    // Note: Space character is percent encoded, rather than as "+" as it's a valid encoding,
+    // and less ambiguous (and we don't need the "+"-encoding for aesthetic reasons)
+    
+    // Starts with the standard URL-allowed character set.
+    NSMutableCharacterSet *allowedCharacters =
+        [[NSCharacterSet URLQueryAllowedCharacterSet] mutableCopy];
+    // Keep only alphanumeric characters.
+    [allowedCharacters formIntersectionWithCharacterSet:[NSCharacterSet alphanumericCharacterSet]];
```
StevenEWright: This pattern would not have occurred to me because I would have assumed that alphanumeric characters are a strict subset of the above.

I like what you have because it’s explicit - but I have a question for my own benefit.... is my assumption incorrect?

Source/OIDTokenRequest.m
```diff
@@ -274,7 +274,24 @@ - (NSURLRequest *)URLRequest {
...
+    
+    // Percent encode all non-alphanumeric characters.
+    NSString *encodedClientID =
+        [_clientID stringByAddingPercentEncodingWithAllowedCharacters: allowedCharacters];
```
StevenEWright: No space after colon for consistency. (And below)

Source/OIDTokenRequest.m
julienbodet: That's what I thought at first but that's not true, alphanumericCharacterSet contains a lot of other alphanumeric characters (thousands of characters!!) that are not part of the US-ASCII characters.

alphanumericCharacterSet from Apple documentation:
A character set containing the characters in Unicode General Categories L*, M*, and N*.

Another way to do would be to call removeCharactersInString: on the first set by specifying special characters. Or defining our own set by including characters  from A...Z, a...z and 0...9.

StevenEWright: Awesome! Thanks for that!!

WilliamDenniss: I think it's a mistake to encode *every* non-alphanumeric character. This results in non-reserved characters like `.` being encoded. Some systems may not expect this to happen.  E.g. let's say I issue clients with alphanumerics, and `.` characters only (like a certain well-known IdP). Since I know I'm not using any reserved characters, my OAuth implementation could be lazy and *not* url-decode these values. Now it will break. 

Maybe we should just follow the [latest spec](https://url.spec.whatwg.org/#urlencoded-serializing) on `application/x-www-form-urlencoded` (linked from [here](https://www.w3.org/TR/html5/sec-forms.html#application-x-www-form-urlencoded-encoding-algorithm), including the treatment of `+`).

That spec defines the following serializer:
```
0x20 (SP)
Append U+002B (+) to output.
0x2A (*)
0x2D (-)
0x2E (.)
0x30 (0) to 0x39 (9)
0x41 (A) to 0x5A (Z)
0x5F (_)
0x61 (a) to 0x7A (z)
Append a code point whose value is byte to output.

Otherwise
Append byte, percent encoded, to output.
```

Implementing this logic in iOS with the charset methods you're using would result in logic like so:
1. Build a charset allowing: ` *-._A-Za-z0-9` (nb. space included). %-encode any char outside this range.
2. Substitute ` ` for `+`.

NB. I withdraw my comment about `+` since it's actually pretty clear on this front, both the W3C, and the OAuth2 spec. Space is encoded as `+`, and `+` gets percent-encoded.

Source/OIDTokenRequest.m
```diff
@@ -274,7 +274,24 @@ - (NSURLRequest *)URLRequest {
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
 
   if (_clientSecret) {
-    NSString *credentials = [NSString stringWithFormat:@"%@:%@", _clientID, _clientSecret];
+    // https://tools.ietf.org/html/rfc6749#appendix-B
```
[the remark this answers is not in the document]
WilliamDenniss: And a link to https://www.w3.org/TR/html5/sec-forms.html#application-x-www-form-urlencoded-encoding-algorithm to explain *how* we chose to encode this.

Source/OIDTokenRequest.m
julienbodet: Thanks for your detailed explanation. To build the character set I followed instructions from https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1 saying:

> Non-alphanumeric characters are replaced by `%HH', a percent sign and two hexadecimal digits representing the ASCII code of the character.

But maybe the definition of alphanumeric here is not the same as on iOS.

StevenEWright: Thinking about this further... Julien, could you add some unit tests to clarify the situation here?

julienbodet: Yes I will!

Source/OIDTokenUtilities.m
```diff
@@ -63,4 +69,21 @@ + (NSString *)redact:(NSString *)inputString {
   }
 }
 
++ (NSString*)formUrlEncode:(NSString*)inputString {
```
julienbodet: Is it the right location to put this method?

WilliamDenniss: Works for me.

Source/OIDTokenRequest.m
```diff
@@ -274,7 +275,12 @@ - (NSURLRequest *)URLRequest {
   NSMutableDictionary *httpHeaders = [[NSMutableDictionary alloc] init];
 
   if (_clientSecret) {
-    NSString *credentials = [NSString stringWithFormat:@"%@:%@", _clientID, _clientSecret];
+    // https://tools.ietf.org/html/rfc6749#section-2.3.1
```
WilliamDenniss: nit: I'd add some text here like 
// form-urlencode the client id and secret per RFC 6749 Section 2.3.1
// https://tools.ietf.org/html/rfc6749#section-2.3.1

StevenEWright:
Thanks, as always, @julienbodet for the awesome contributions! 🏅

PR #268, opened 2018-07-11
Add request trace logging enabled via the preprocessor define

WilliamDenniss:
Useful when debugging IdP issues.


gretzki:
@jerneyio You have to put that statement under precompiler macros under the pods / appauth target.

PR #273, opened 2018-07-14
Allow user to specify dispatch queue for the `performActionWithFreshTokens` method

### LINKED ISSUE #46: Dispatching callback/completion/action blocks on the main queue is not always appropriate
There are a few methods where callback blocks are dispatched on the main queue:

```
+ [OIDAuthorizationService discoverServiceConfigurationForDiscoveryURL:completion:]
+ [OIDAuthorizationService performTokenRequest:callback:]
- [OIDAuthState withFreshTokensPerformAction:]
```

This strategy makes sense if the callback block is expected to directly perform UI updates, but this will not always be the case. In cases where the callback blocks instead perform significant computation, this behavior is generally undesirable and forces calling code to perform additional dispatching.

I think it would be ideal if a mechanism could be provided to specify the queue for callback dispatch. For possible approaches, take a look at the mechanisms provided by NSURLSession and Alamofire.

WilliamDenniss:
To date, all AppAuth blocks execute on the main queue. This allows the user to specify the queue for `performActionWithFreshTokens`. Unlike the rest of AppAuth's blocks, this one is called frequently – for every API call – thus it is important for the user to determine on which dispatch queue it executes.

Resolves #46


Source/OIDAuthState.m
```diff
@@ -477,7 +487,7 @@ - (void)performActionWithFreshTokens:(OIDAuthStateAction)action
                  originalAuthorizationResponse:_lastAuthorizationResponse
                                       callback:^(OIDTokenResponse *_Nullable response,
                                                  NSError *_Nullable error) {
-    dispatch_async(dispatch_get_main_queue(), ^() {
+    dispatch_async(dispatchQueue, ^() {
```
WilliamDenniss: I'm thinking that the lines updating OIDAuthState should be in dispatch_get_main_queue(), and the "actionToProcess" block below in the user's dispatchQueue, do you agree @StevenEWright?

PR #275, opened 2018-07-19
Perform case-insensitive comparison of scheme during callback validation

thatjuan:
When verifying that a callback url matches what's expected, the scheme is compared case-sensitively. 

As of iOS11, the SFAuthenticationSession process passes in the callback with an always all-lowercase scheme. If we've added a mixed-case scheme, this comparison will fail silently and the callback will not be handled; Causing the login process to hang.

[RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.1) calls for schemes to be case-insensitive. For that reason, performing a case-sensitive check on the scheme with `OIDIsEqualIncludingNil` is not necessary, and, in this case, incorrect.

I've also changed the structure of the check a bit for readability. However, this could easily be re-written as an all-in-one return the way it was before.

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -96,12 +96,20 @@ NS_ASSUME_NONNULL_BEGIN
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [_request.redirectURL standardizedURL];
 
-  return OIDIsEqualIncludingNil(standardizedURL.scheme, standardizedRedirectURL.scheme) &&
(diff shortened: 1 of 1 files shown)


WilliamDenniss:
@thatjuan thank you for the PR. Can you please sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/) so we can consider your pull request?

WilliamDenniss:
Hi @thatjuan, do you plan to sign the CLA?

WilliamDenniss:
Hi @thatjuan, any luck with the CLA?

WilliamDenniss:
Closing in favor of #398

We don't need a nil check on the scheme.

Issue #282, opened 2018-09-03
Make nonce optional for OAuth servers that don't support it.


FelipeBuiles:
AppAuth-iOS worked correctly with [AWS Cognito](https://aws.amazon.com/cognito/) up until its 0.93 release that fixed #4. The issue lies in the inabilty of Cognito to handle the nonce parameter (mentioned in their forums [here](https://forums.aws.amazon.com/thread.jspa?messageID=802423&tstart=0)).
There is a comment on the `OIDAuthorizationRequest.h` file that mentions something about it being nullable:
```
    @param nonce String value used to associate a Client session with an ID Token. Can be set to nil
        if not using OpenID Connect, although pure OAuth servers should ignore params they don't
        understand anyway.
```
The thing is, my team is not using AppAuth-iOS directly but rather trough a [wrapper](https://github.com/FormidableLabs/react-native-app-auth) since our project is built with React Native, and we need a way to expose this functionality to that wrapper.

I've opened a [PR](https://github.com/openid/AppAuth-iOS/pull/283) that implements our fix to this, its also linked to another PR on the wrapper repo.

WilliamDenniss:
The right approach here (to avoid sending nonce to a non-compliant provider) is to use the designated initializer of OIDAuthorizationRequest and pass `nil` for the nonce.

https://github.com/openid/AppAuth-iOS/blob/b4ca39a38f0d991bd17c56ebd01756a41da54061/Source/OIDAuthorizationRequest.h#L183-L216

FelipeBuiles:
You're completely right, I'll update my PR on their repo to use the correct initializer. Thanks for the guidance.

PR #283, opened 2018-09-03
Expose a useNonce parameter on initWithConfiguration to make nonce nullable.

FelipeBuiles:
Implement the solution described in #282


WilliamDenniss:
I don't think this is needed. We already support not sending nonce. You can pass `nil` into the designated initializer:

https://github.com/openid/AppAuth-iOS/blob/b4ca39a38f0d991bd17c56ebd01756a41da54061/Source/OIDAuthorizationRequest.h#L183-L216

Nonce is not added if nil:

https://github.com/openid/AppAuth-iOS/blob/b4ca39a38f0d991bd17c56ebd01756a41da54061/Source/OIDAuthorizationRequest.m#L319-L321

PR #287, opened 2018-09-13
Fix typo (sha265 -> sha256)

jogu:
Trivial but it gave me pause for thought so probably worth fixing.

I've signed a CLA: https://openid.net/wordpress-content/uploads/2018/04/OIDF_CLA-Agreement-Individual_Joseph-Heenan_2018-03-30.pdf


WilliamDenniss:
Thanks for catching this!

Issue #105, opened 2017-04-03
Regression: Exception when deserializing migrated GTMOAuth2 objects


chrismcs:
This is due to a recent change made in #99 

I'm relatively new to AppAuth and GTMAppAuth, so it's very possible I'm misunderstanding how something should be working here - but here's what I'm seeing.

As part of #99 - if you now attempt to init an `OIDAuthorizationRequest` with a response type of `token`, an exception is thrown.  The problem is that GMTOAuth2 apparently used to use `token` as a response type, therefore a couple (related) things now break:

1.  An exception is thrown when merely deserializing a stored OIDAuthorizationRequest that was set to `token` as the response type

2.  https://github.com/google/GTMAppAuth - in the `GTMOAuth2KeychainCompatibility.m`, when they are deserializing the old keychain entry to migrate it to a new one, they explicitly use `token` as the response type in `authorizeFromPersistenceString:serviceProvider:tokenURL:redirectURI:clientID:clientSecret:`.  

3.  Even after I do the migration (by commenting out the change in #99), the problem seems to be that `OIDAuthState` stores a request/response history, so `_lastAuthorizationResponse` is set to be a `token` response.  So, if I make the changes for #99 live again, it will still throw the exception every time it grabs the new authorization object that is stored in the keychain.

So, should the changes in #99 be modified to not crash when deserializing an `OIDAuthorizationRequest` or is there some other way to handle this?

mike-ferenduros:
+1
This makes the recommended migration path to app auth unusable.

StevenEWright:
I'm just seeing this now. Wow. We'll get right on this.

StevenEWright:
Also, @chrismcs thank you for the excellent problem summary.

StevenEWright:
@WilliamDenniss I sat down to work on this, and realized it's not as simple as "we made a mistake". Can you comment on the progeny of this issue? It actually looks like this is WAI?

See: https://github.com/openid/AppAuth-iOS/issues/98

StevenEWright:
The related error in GTMAppAuth is: https://github.com/google/GTMAppAuth/issues/35

StevenEWright:
Thanks for the fast review, @WilliamDenniss 
Thanks for the problem report, @chrismcs

PR #298, opened 2018-09-20
Allow applications to use the "code id_token" response_type

### LINKED ISSUE #292: Very awkward for developers using appauth to use hybrid flow
**Describe the bug**

It's very awkward for library users to use hybrid flow.

**To Reproduce**

If you set the response_type to code id_token, appauth throws an assert:

https://github.com/openid/AppAuth-iOS/blob/master/Source/OIDAuthorizationRequest.m#L134

which references:

https://github.com/openid/AppAuth-iOS/issues/98

which says:

> After discussing id_token support with @ve7jtb, we agreed that it could be possible to use code id_token in a secure way with AppAuth, if you validate the ID Token, verify the c_hash, and only use the ID Token after the code is exchanged successfully with PKCE verification.

I agree with this comment. I've implemented id token signature verification and c_hash verification in my app in exactly this way.

and:

> It's not currently supported and there are no plans to do so. Hybrid (and implicit) negate the PKCE protection required by RFC 8252 so it's generally not recommended for native apps.

I'm not sure I understand this comment (you can use hybrid and PKCE together, and FAPI part 2 requires this for public clients.)

The actual assert was added as part of https://github.com/openid/AppAuth-iOS/pull/117

**Expected behavior**

AppAuth should allow library users to use the hybrid flow as library users can implement id_token and c_hash verification. As far as I can see it's currently either awkward or impossible to do this without changing the AppAuth source code, as this assert is in the designated initialiser for OIDAuthorizationRequest.

**Additional context**

FAPI requires the use of the hybrid flow, so it's not currently possibly to make a FAPI compliant OIDC client without modifying appauth.

jogu:
This response type is required to build a FAPI compliant client, and as
the application can verify the id_token signature and c_hash before
calling the token endpoint there is no reason to block applications from
using code id_token.

Other response types are still blocked as they don't appear to make sense
in the context of a native app.

closes #292

### DIFF
--- a/Source/OIDAuthState.h
+++ b/Source/OIDAuthState.h
@@ -117,7 +117,11 @@ typedef void (^OIDAuthStateAuthorizationCallback)(OIDAuthState *_Nullable authSt
 @property(nonatomic, weak, nullable) id<OIDAuthStateErrorDelegate> errorDelegate;
 
 /*! @brief Convenience method to create a @c OIDAuthState by presenting an authorization request
-        and performing the authorization code exchange in the case of code flow requests.
+        and performing the authorization code exchange in the case of code flow requests. For
+        the hybrid flow, the caller should validate the id_token and c_hash, then perform the token
+        request (@c OIDAuthorizationService.performTokenRequest:callback:)
+        and update the OIDAuthState with the results (@c
+        OIDAuthState.updateWithTokenResponse:error:).
     @param authorizationRequest The authorization request to present.
     @param externalUserAgent A external user agent that can present an external user-agent request.
     @param callback The method called when the request has completed or failed.
--- a/Source/OIDAuthState.m
+++ b/Source/OIDAuthState.m
@@ -159,8 +159,13 @@ static const NSUInteger kExpiryTimeTolerance = 60;
                                                 callback(authState, tokenError);
                                }];
                              } else {
-                               // implicit or hybrid flow (hybrid flow assumes code is not for this
-                               // client)
+                               // hybrid flow (code id_token). Two possible cases:
+                               // 1. The code is not for this client, ie. will be sent to a
+                               //    webservice that performs the id token verification and token
+                               //    exchange
+                               // 2. The code is for this client and, for security reasons, the
+                               //    application developer must verify the id_token signature and
+                               //    c_hash before calling the token endpoint
                                OIDAuthState *authState = [[OIDAuthState alloc]
                                    initWithAuthorizationResponse:authorizationResponse];
                                callback(authState, authorizationError);
--- a/Source/OIDAuthorizationRequest.m
+++ b/Source/OIDAuthorizationRequest.m
@@ -87,7 +87,7 @@ static NSUInteger const kCodeVerifierBytes = 32;
 /*! @brief Assertion text for unsupported response types.
  */
 static NSString *const OIDOAuthUnsupportedResponseTypeMessage =
-    @"The response_type \"%@\" isn't supported. AppAuth only supports the \"code\" response_type.";
+    @"The response_type \"%@\" isn't supported. AppAuth only supports the \"code\" or \"code id_token\" response_type.";
 
 /*! @brief Code challenge request method.
  */
@@ -104,6 +104,23 @@ NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256 = @"S256";
                            responseType:
                    additionalParameters:)
     )
+
+/*! @brief Check if the response type is one AppAuth supports
+    @remarks AppAuth only supports the `code` and `code id_token` response types.
+    @see https://github.com/openid/AppAuth-iOS/issues/98
+    @see https://github.com/openid/AppAuth-iOS/issues/292
+ */
++ (BOOL)isSupportedResponseType:(NSString *)responseType
+{
+  NSString *codeIdToken = [@[OIDResponseTypeCode, OIDResponseTypeIDToken]
+                           componentsJoinedByString:@" "];
+  NSString *idTokenCode = [@[OIDResponseTypeIDToken, OIDResponseTypeCode]
+                           componentsJoinedByString:@" "];
+
+  return [responseType isEqualToString:OIDResponseTypeCode]
+         || [responseType isEqualToString:codeIdToken]
+         || [responseType isEqualToString:idTokenCode];
+}
 
 - (instancetype)initWithConfiguration:(OIDServiceConfiguration *)configuration
                 clientId:(NSString *)clientID
@@ -126,11 +143,7 @@ NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256 = @"S256";
     _scope = [scope copy];
     _redirectURL = [redirectURL copy];
     _responseType = [responseType copy];
-    // Attention: Please refer to https://github.com/openid/AppAuth-iOS/issues/105
-    // If you change the restriction on response type here, you must also update initWithCoder:
-    if (![_responseType isEqualToString:OIDResponseTypeCode]) {
-      // AppAuth only supports the `code` response type.
-      // Discussion: https://github.com/openid/AppAuth-iOS/issues/98
+    if (![[self class] isSupportedResponseType:_responseType]) {
       NSAssert(NO, OIDOAuthUnsupportedResponseTypeMessage, _responseType);
       return nil;
     }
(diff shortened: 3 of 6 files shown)


jogu:
@StevenEWright  I considered whether this re-introduces https://github.com/openid/AppAuth-iOS/issues/105 as it effectively undoes part of your fix for that issue.

My belief is that as no exception is thrown a release build of the app won't crash and hence there's not an issue. Deserialising a broken object (one where the stored response type was not code) will return nil, but presumably apps cope with that without crashing.

We could possibly maintain the current behaviour (where I believe deserialised broken objects forces the response type to be code) by changing initWithCoder to treat any unsupported response type as code - I'm reluctant to do that unless anyone thinks it's still necessary.

WilliamDenniss:
> @StevenEWright I considered whether this re-introduces #105 as it effectively undoes part of your fix for that issue.

Thanks for pointing this out.

Note that root GTMAppAuth bug was fixed in https://github.com/google/GTMAppAuth/commit/cbd204dd98a8af75e14644b7c635c59ed52df587 some time ago, so hopefully there are not a lot of serialized objects out there.

> 
> My belief is that as no exception is thrown a release build of the app won't crash and hence there's not an issue. Deserialising a broken object (one where the stored response type was not code) will return nil, but presumably apps cope with that without crashing.

I'm inclined to agree, now that it's an assert, it should return `nil` in release builds which would hopefully just cause a re-auth.

> 
> We could possibly maintain the current behaviour (where I believe deserialised broken objects forces the response type to be code) by changing initWithCoder to treat any unsupported response type as code - I'm reluctant to do that unless anyone thinks it's still necessary.

+1 to not do this.

Source/OIDAuthState.m
```diff
@@ -159,8 +159,13 @@ @implementation OIDAuthState {
...
                              } else {
-                               // implicit or hybrid flow (hybrid flow assumes code is not for this
-                               // client)
+                               // hybrid flow (code id_token). Two possible cases:
+                               // 1. The code is not for this client, ie. will be sent to a
+                               //    webservice that performs the id token verification and token
+                               //    exchange
+                               // 2. The code is for this client and, for security reasons, the
+                               //    application developer must verify the id_token signature and
+                               //    c_hash before calling the token endpoint
```
WilliamDenniss: add a note perhaps… "for that reason, the code exchange is not performed automatically" ?

Source/OIDAuthorizationRequest.m
```diff
@@ -105,6 +105,20 @@ - (instancetype)init
...
     )
 
+/*! @brief Check if the response type is one AppAuth supports
+    @remarks AppAuth only supports the `code` and `code id_token` response types.
+    @see https://github.com/openid/AppAuth-iOS/issues/98
+    @see https://github.com/openid/AppAuth-iOS/issues/292
+ */
++ (BOOL)isSupportedResponseType:(NSString *)responseType
+{
+  NSString *codeIdToken = [@[OIDResponseTypeCode, OIDResponseTypeIDToken]
```
WilliamDenniss: Should we accept `id_token code` as well? I don't know if the order matters, is it documented?

Source/OIDAuthorizationRequest.m
```diff
@@ -105,6 +105,20 @@ - (instancetype)init
...
+      componentsJoinedByString:@" "];
+
+  return [responseType isEqualToString:OIDResponseTypeCode] ||
```
WilliamDenniss: Put the `||` on the next line, aligned with the `[`

Source/OIDAuthState.m
WilliamDenniss: I assume that for hybrid you won't use this convenience initializer, is that how you've implemented it?

Source/OIDAuthorizationRequest.m
jogu: Good question. It looks like https://tools.ietf.org/html/rfc6749#section-3.1.1 applies, which says:

>  Extension response types MAY contain a space-delimited (%x20) list of values, where the order of values does not matter (e.g., response type "a b" is the same as "b a").

So yes, we probably should cope with that.

Source/OIDAuthState.m
jogu: We actually do use this convenience initialiser, and pick up the id_token up from authState.lastAuthorizationResponse.idToken - I can't say that was a deliberate decision though. Does using it for the hybrid flow worry you in any way?

WilliamDenniss: Not it doesn't worry me, but I think we should be more explicit that hybrid is officially "supported" by this method.

Can you enhance the method doc? Currently it reads:

> Convenience method to create a @c OIDAuthState by presenting an authorization request
        and performing the authorization code exchange in the case of code flow requests.

It doesn't exclude hybrid but doesn't exactly tell the user much either. Documenting the behavior for hybrid would be good (i.e. it doesn't do the code exchange, you're expected to validate the token then do the exchange yourself, and call the updateWithTokenResponse method).

Source/OIDAuthorizationRequest.m
jogu: Fixed.

Source/OIDAuthState.m
jogu: Yes. That makes sense. I'll push this change shortly, thanks.

Source/OIDAuthorizationRequest.m
jogu: I've made this change.

Just to be thorough I also checked on whether multiple spaces are allowed between the values, and by my reading the RFC is explicit that only a single space is permitted.

jogu:
Thanks for the quick review - I've made all the requested changes, so hopefully this is good to go.

I'm not sure if this project has a preference for squashing review changes down into the original commit or putting them as new commits; I've gone for the latter - if you'd like me to squash them let me know.

Source/OIDAuthState.h
```diff
@@ -117,7 +117,9 @@ typedef void (^OIDAuthStateAuthorizationCallback)(OIDAuthState *_Nullable authSt
 @property(nonatomic, weak, nullable) id<OIDAuthStateErrorDelegate> errorDelegate;
 
 /*! @brief Convenience method to create a @c OIDAuthState by presenting an authorization request
-        and performing the authorization code exchange in the case of code flow requests.
+        and performing the authorization code exchange in the case of code flow requests. For
+        the hybrid flow, the caller should validate the id_token and c_hash, then call
+        [OIDAuthorizationService performTokenRequest] then [OIDAuthState updateWithTokenResponse]
```
WilliamDenniss: … c_hash, then perform the token request (@c OIDAuthorizationService.performTokenRequest:callback:)
and updating the OIDAuthState with the results (@c OIDAuthState.updateWithTokenResponse:error:).

Source/iOS/OIDAuthState+IOS.h
```diff
@@ -27,7 +27,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface OIDAuthState (IOS)
 
 /*! @brief Convenience method to create a @c OIDAuthState by presenting an authorization request
-        and performing the authorization code exchange in the case of code flow requests.
+        and performing the authorization code exchange in the case of code flow requests. For
+        the hybrid flow, the caller should validate the id_token and c_hash, then call
+        [OIDAuthorizationService performTokenRequest] then [OIDAuthState updateWithTokenResponse]
```
WilliamDenniss: as above

Source/macOS/OIDAuthState+Mac.h
```diff
@@ -25,7 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface OIDAuthState (Mac)
 
 /*! @brief Convenience method to create a @c OIDAuthState by presenting an authorization request
-        and performing the authorization code exchange in the case of code flow requests.
+        and performing the authorization code exchange in the case of code flow requests. For
+        the hybrid flow, the caller should validate the id_token and c_hash, then call
+        [OIDAuthorizationService performTokenRequest] then [OIDAuthState updateWithTokenResponse]
```
WilliamDenniss: as above

Source/OIDAuthorizationRequest.m
```diff
@@ -126,11 +143,7 @@ - (instancetype)initWithConfiguration:(OIDServiceConfiguration *)configuration
     _scope = [scope copy];
     _redirectURL = [redirectURL copy];
     _responseType = [responseType copy];
-    // Attention: Please refer to https://github.com/openid/AppAuth-iOS/issues/105
-    // If you change the restriction on response type here, you must also update initWithCoder:
-    if (![_responseType isEqualToString:OIDResponseTypeCode]) {
-      // AppAuth only supports the `code` response type.
-      // Discussion: https://github.com/openid/AppAuth-iOS/issues/98
+    if (![self.class isSupportedResponseType:_responseType]) {
```
WilliamDenniss: Only properties should use dot notation per style guide `[self class]`, sorry I missed this earlier.

Other than that it's ready :)

PR #303, opened 2018-09-24
Update minimum supported iOS version to 8.0

jogu:
AppAuth now requires Xcode 10 / iOS 12 SDK, and hence does not support
iOS 7 anymore.

Update the minimum version in all the places I can find it and update
the documentation.


jogu:
Actually I now see from https://github.com/openid/AppAuth-iOS/commit/2c934c924de5a294fb64b7460e31cba2fb005aac that this seems to be deliberate:

> AppAuth now requires Xcode 10 to build (but still supports a deployment target of iOS 7), per our policy to only support the current development tools.

I'm not sure if I'm missing something but I can't see how this can be the case. If you build with Xcode 10 with the iOS deployment target set to 7.0 it says:

> warning: The iOS deployment target is set to 7.0, but the range of supported deployment target versions for this platform is 8.0 to 12.0. (in target 'AppAuth-iOS')

I can't currently think of any sort of even semi-supported way that might let you build the current version of AppAuth in a way that running it on iOS 7 was supported.

jogu:
Good question.

The Xcode 10 release notes say:

> Xcode 10 is available in the Mac App Store and includes SDKs for iOS 12, watchOS 5, macOS 10.14, and tvOS 12. Xcode 10 supports on-device debugging for iOS 8 and later, tvOS 9 and later, and watchOS 2 and later. Xcode 10 requires a Mac running macOS 10.13.6 or later.

Xcode itself also doesn't appear to support installing a simulator for anything other than iOS 8.1, so I presume debugging on an earlier simulator is also impossible/unsupported.

So it would appear to me that it might be technically possible do as you suggest but (as I believe it is unsupported to link a static binary built for the iOS 12 SDK with an earlier iOS SDK, ie. the one in Xcode 9.4) you would never be able to debug the application - at least not without resorting to other unsupported things like copying SDKs or simulators from one version of Xcode into a different version, or using a laborious ad-hoc install & NSLog style of debugging

So in summary I'd tend to think that whilst you're right that it may well be possible, on balance  if there's a strong desire to continue to support iOS 7 then it would be significantly simpler to also continue to support Xcode 9.4. My opinion is that supporting iOS 7 probably isn't that important (if someone really needs it, which is only likely to be the case in some unusual off-appstore app development targeting old cheap devices, just use an older version of AppAuth).

(In my iOS consultant role, I've always found it easy to persuade clients to drop iOS versions not supported by Xcode so I'd never had to look at this in that much detail before. In fact they generally prefer to not even go that far back given the testing overhead it adds whilst only adding a very small number of users.)

WilliamDenniss:
I don't have a strong desire to continue supporting iOS 7 (we've carried it since iOS 9 was the latest version, so I think we've done pretty well), but I'm also not in a rush to remove it for as long as it's usable and not much of a maintenance burden.

Regarding usability, my assumption due to the lack of simulator support was that you could still debug on real hardware (I don't know if this is actually the case, however). Anyone still targeting 7.0 presumably should have hardware available, and I've worked on teams that do maintain hardware going back several versions.

For maintenance, the ongoing support cost to us appears to be minimal. We're already maintaining all the iOS 7 code for a combination of macOS 10.9, and iOS 8 anyway, and I've not seen any iOS 7 specific bugs (which could be because nobody is using it too, of course).

I'd probably vote to drop it when one of the following happens:

1. iOS 7 can't be targeted by the current Xcode tool chain.
2. The AppAuth support for iOS 7 is broken, and we can't debug it ourselves.
3. We can remove a decent amount of code due to dropping support.

Generally speaking, our goal with the library is to support as many versions as possible, since we're a core dependency and our minimum pushes up the minimums of anyone who consumes the library (which is a different calculation then when considering what version a end-product app should target – I'm with you there, and normally advocate for only 1-2 generations of version support myself, depending on how widely deployed the app is).

In any case, this PR is good for when the time comes – and if you think that time should be now, I'm open to being convinced.

WilliamDenniss:
Closing this for now, but the code is good – can re-open and use in the future when the need arises.

PR #310, opened 2018-10-08
Create subpods 'Core' and 'Auth'

schroepf:
Fix bug #198 by implementing the design suggested in [a comment in PR 218](https://github.com/openid/AppAuth-iOS/pull/218#issuecomment-427861770) this allows Cocoapods users to depend on:

- `AppAuth` if they need the fully blown AppAuth.framework (including platform specific APIs)
- `AppAuth/Core` if they cannot use platform specific APIs (like when running in an iOS App Extension)


julienbodet:
Great! I think you should also define a 'Core' framework target (and static library?) for those importing AppAuth via Carthage. Otherwise they still won't be able to import AppAuth in extensions.

WilliamDenniss:
Can you sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/), or if you have post a link to the PDF?

schroepf:
Yes, sure: [Signed CLA](https://www.docusign.net/Member/EmailStart.aspx?a=2954cc73-d213-4d22-a6ac-e796dc1dd613&r=81eac43f-ad07-46f6-a565-28ed0a369ee7)

AppAuth.podspec
```diff
@@ -33,16 +33,21 @@ It follows the OAuth 2.0 for Native Apps best current practice
...
-  s.ios.source_files      = "Source/iOS/**/*.{h,m}"
-  s.ios.deployment_target = "7.0"
-  s.ios.frameworks        = "SafariServices", "AuthenticationServices"
-
-  # macOS
-  s.osx.source_files = "Source/macOS/**/*.{h,m}"
-  s.osx.deployment_target = '10.9'
+  s.subspec 'Core' do |core|
+     # subspec for users who don't want the authentication code
+     core.source_files = "Source/*.{h,m}"
```
WilliamDenniss: Can this line be moved outside the subspec?

AppAuth.podspec
```diff
@@ -33,16 +33,21 @@ It follows the OAuth 2.0 for Native Apps best current practice
   s.platforms    = { :ios => "7.0", :osx => "10.9", :watchos => "2.0", :tvos => "9.0" }
 
   s.source       = { :git => "https://github.com/openid/AppAuth-iOS.git", :tag => s.version }
-
-  s.source_files = "Source/*.{h,m}"
```
WilliamDenniss: Suggest reverting this (see below).

AppAuth.podspec
```diff
@@ -33,16 +33,21 @@ It follows the OAuth 2.0 for Native Apps best current practice
...
-
-  # macOS
-  s.osx.source_files = "Source/macOS/**/*.{h,m}"
-  s.osx.deployment_target = '10.9'
+  s.subspec 'Core' do |core|
+     # subspec for users who don't want the authentication code
+     core.source_files = "Source/*.{h,m}"
+  end
+
+  s.subspec 'Auth' do |auth|
```
WilliamDenniss: Let's call this "ExternalUserAgent" since that's what the functionality adds, and "AppAuth/Auth" has too many Auths in it :)

AppAuth.podspec
```diff
@@ -33,16 +33,21 @@ It follows the OAuth 2.0 for Native Apps best current practice
...
-  # iOS
-  s.ios.source_files      = "Source/iOS/**/*.{h,m}"
-  s.ios.deployment_target = "7.0"
-  s.ios.frameworks        = "SafariServices", "AuthenticationServices"
-
-  # macOS
-  s.osx.source_files = "Source/macOS/**/*.{h,m}"
-  s.osx.deployment_target = '10.9'
+  s.subspec 'Core' do |core|
+     # subspec for users who don't want the authentication code
```
WilliamDenniss: "subspec for users who don't need the external user agent implementations"

and move to above the s.subspec line.

Source/AppAuth.h
```diff
@@ -1,90 +0,0 @@
-/*! @file AppAuth.h
```
schroepf: Ok, I think I got the reason why this `AppAuth.h` was there (and why it's a bad idea to delete it... ;) ):
- it's not for Carthage (carthage uses the still existing `Source/Framework/AppAuth.h` (with all the `#import <AppAuth/...>` style imports
- this header file IS used by cocoapods users, but I didn't notice as I was testing with the `Example-iOS_ObjC` sample project which imported the `AppAuth.h` right via the "Header Search Path"  (which is set to "../../Source/*") and not the `<AppAuth/AppAuth.h>` header which should have been bundled with the pod
- the `Example-macOS` project uses the correct import an actual app would probably also use:
  ```
  #import <AppAuth/AppAuth.h>
  ```
  and this example project fails to build due to the missing `AppAuth.h` file

AppAuth.podspec
schroepf: done

AppAuth.podspec
schroepf: yes, I agree... I also thought about better alternatives to "Auth" (like "Native" or "Platform") but they didn't really fit... so yes... "ExternalUserAgent" it is...

AppAuth.podspec
schroepf: the reason why I had to put it inside the subspec is the following cocoapods error:

```
[!] The `AppAuth` pod failed to validate due to 1 error:
    - ERROR | File Patterns: The AppAuth/Core (0.94.0) spec is empty (no source files, resources, resource_bundles, preserve paths, vendored_libraries, vendored_frameworks, dependencies, nor subspecs).
```

so as long as the subspec remains empty cocoapods refuses to build the pod...

AppAuth.podspec
schroepf: due to the cocoapods error for empty subspecs (refer to my other comment) I had to add at least something to the subspec, so I decided to move the `source_files` into the `Core` subspec...

Source/AppAuth.h
schroepf: After a lot of trial and error I thought I found a solution for the header problem:
Restore the `AppAuth.h` and split its contents up into: `AppAuthCore.h`, `iOS/AppAuth.h` and `macOS/AppAuth.h` (and have the two `AppAuth.h` files import `AppAuthCore.h`. All the Example and all of my test projects could deal with that solution via a CocoaPods dependency.

But as it turned out having two AppAuth.h files in the project breaks the .framework builds and the CI is failing with my last commit.

I also tried to implement the proposed solution from [this comment](https://github.com/openid/AppAuth-iOS/issues/198#issuecomment-428158360) and tried to use a preprocessor macro and `#ifdef` statements to fix the original header file but couldn't find a way to set the preprocessor macro depending on the subspec from the podspec.

And the longer I think about it, and considering [this comment](https://github.com/openid/AppAuth-iOS/pull/310#issuecomment-428017038) it appears to me that maybe the usage of subspecs isn't the correct way to implement this. And maybe the original approach to create a separate podspec for the core framework (+ separate build targets to build a core framework via carthage) would be a better way to solve the issues...

AppAuth.podspec
schroepf: `AppAuth/Core` will work, because that's the subspec which includes all the files in `Source/*.{h,m}` ;-)

But those who include `AppAuth/ExternalUserAgent` explicitly will also work, because the `ExternalUserAgent` subspec has specified `AppAuth/Core` as its dependency

Source/AppAuth.h
julienbodet: > I also tried to implement the proposed solution from this comment and tried to use a preprocessor macro and #ifdef statements to fix the original header file but couldn't find a way to set the preprocessor macro depending on the subspec from the podspec.

What did you try to set the preprocessor macro depending on the subspec?

Did you try adding the following to the `AppAuth/ExternalUserAgent` subspec?

`externalUserAgent.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -MY_PREPROCESSOR_MACRO' }`

as explained [here](http://www.dbotha.com/2014/12/04/optional-cocoapod-dependencies/).

schroepf: to be honest no... I think I used `GCC_PREPROCESSOR_DEFINITIONS` instead of `OTHER_CFLAGS`.... I'll give it one more try...

Source/AppAuth.h
```diff
@@ -41,6 +41,8 @@
 #import "OIDTokenUtilities.h"
 #import "OIDURLSessionProvider.h"
 
+#ifdef APPAUTH_USERAGENT
```
WilliamDenniss: This will break existing users who are not using Pods, needs to be inverted (#ifndef _NO_APPAUTH_USERAGENT and define it in Core).

AppAuth.podspec
WilliamDenniss: Ah ok, sounds good.

Source/AppAuth.h
julienbodet: I don't think it'll work by inverting this because Core is now included as a dependency in ExternalUserAgent, so I guess this macro will always be defined... And also if no spec is specified in the user podfile, I think all specs will be included.

If this is confirmed, maybe another solution is to have ExternalUserAgent as a default_subspec and include all the files under Source folder in it.

I'll try to test it.

WilliamDenniss: Good points, look forward to hearing the results of your experiment!

julienbodet: I confirm that the flag is always defined no matter which subspec is specified in the Podfile. And I also had another problem when importing `pod AppAuth/Core` in a Swift project, #ifndef seems to be ignored and iOS specific files/types are imported...

schroepf: I didn't have much time to test this yet, but I also couldn't make it work with defining `default_subspec` and inverting the preprocessor macro (i.e. using `#ifndef NO_APPAUTH_USERAGENT`) in the header...

But I'm still thinking if it's really a problem to keep the existing logic (i.e. `#ifdef APPAUTH_USERAGENT`).... in my current understanding the `AppAuth.h` is currently used exclusively for the CocoaPods build. Carthage/manual framework builds use another AppAuth.h header. So do we really need to consider users who are not using Pods in this header file?

WilliamDenniss: If they're using a different header, then I guess it should be fine to keep the existing logic, assuming this file is only used in the Cocoapods environment, where this will always be defined.

Do all the samples (both Cocoapods and Carthage) compile with this PR as it is currently?

@julienbodet can you review & LGTM when/if you confirm that this PR looks good for your use-cases?

schroepf: I guess I need to do some more testing. Although the examples seem to build just fine I'm having issues (in my own project) that at runtime the user agent extension methods of cannot be found:

```
+[OIDAuthState authStateByPresentingAuthorizationRequest:presentingViewController:callback:]: unrecognized selector sent to class 0x10cf7c9f0
```

schroepf:
@WilliamDenniss  @julienbodet : I created an alternative pull request (#313)  which doesn't use cocoapod's subspec feature but introduces a dedicated `AppAuthCore` podspec (similar to what `RxSwift` does). So far I couldn't find any issues with this approach. And what I like about it is that we can implement a consistent solution for carthage (or plain framework) users just by creating an additional `AppAuthCore.framework`.

Source/AppAuth.h
schroepf: Please let me know if calling the category functions works on your side .. but I couldn't make it work for me yet. Seems like the `-ObjC` flag isn't working for subpods or something like that.

WilliamDenniss: Couple of things: 1) I think the headers should match, ideally, divergence just confuses things – and Carthage users probably need this too! 2) The other way to retain the #if logic is to add that define to the xcodeproj build settings so that Carthage users will have it defined, that way they won't break when they upgrade either.

I've gone an implemented this in #318, and I added a new framework target so Carthage users can build with extensions too.

In my testing it works, note that I defined the variable in `GCC_PREPROCESSOR_DEFINITIONS` rather than `OTHER_CFLAGS` in this PR.

schroepf:
I gueass we can close this PR as it's superseded by #320

WilliamDenniss:
Sure thing. Thanks for all your work on this, I've added you to the CONTRIBUTORS acknowledgement file. Good team effort.

PR #313, opened 2018-10-13
Introduce AppAuthCore podspec

schroepf:
Fix bug #198 by implementing a dedicated `AppAuthCore.podspec` which includes only core AppAuth classes but no platform specific user agent code. Cocoapods users can depend on:

- `AppAuth` pod if they need the fully blown AppAuth.framework (including platform specific APIs)
- `AppAuthCore` pod if they cannot use platform specific APIs (like when running in an iOS App Extension)


WilliamDenniss:
Thanks for working on this variant. I'd rather not maintain more podspecs if we don't need to, as it adds more maintenance overhead. Let's try and solve this with subspecs, since this is precisely what subspecs are designed for.

PR #315, opened 2018-10-15
Fixes NSSecureCoding (breaks backward compatibly)

diejmon:
I found quite interesting bug in `OIDServiceDiscovery` `NSSecureCoding` implementation:
if `discoveryDictionary` contains `NSArray`s as values, secure unarchive fails with exception: ```value for key 'NS.objects' was of unexpected class NSArray. Allowed classes are {(
    OIDServiceDiscovery
)}```.

I found a way to fix this, **but** this breaks all current users archives. 

The question is what can we do about this? I can think of the following variants:
- Keep code as is. But this breaks contract, it says that `NSSecureCoding` is supported, while it is not. 
- Ask all users to migrate archives. Not an option in my opinion.
- Modify `initWithCoder:` in a way that will be able to decode old archives, but the new ones will be created with new secure coding implementation. 

Below is possible variant of 3rd option:

```ObjC
- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
  NSError *error;
  __block NSDictionary *dictionary;
  void (^oldImpl)(void) = ^{
    dictionary = [[NSDictionary alloc] initWithCoder:aDecoder];
  };

  @try {
    dictionary = [aDecoder decodeObjectOfClasses:[NSSet setWithObjects:[NSArray class], [NSDictionary class], nil] forKey:kDiscoveryDictionaryKey];
    if (!dictionary) {
      oldImpl();
    }
  }
  @catch (NSException *exception) {
    oldImpl();
  }

  self = [self initWithDictionary:dictionary error:&error];
  if (error) {
    return nil;
  }
  return self;
}
```

### DIFF
--- a/Source/OIDServiceDiscovery.m
+++ b/Source/OIDServiceDiscovery.m
@@ -23,6 +23,10 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/*! @brief Key used to encode the @c discoveryDictionary property for @c NSSecureCoding
+ */
+static NSString *const kDiscoveryDictionaryKey = @"discovery_dictionary";
+
 /*! Field keys associated with an OpenID Connect Discovery Document. */
 static NSString *const kIssuerKey = @"issuer";
 static NSString *const kAuthorizationEndpointKey = @"authorization_endpoint";
@@ -183,7 +187,8 @@ static NSString *const kOPTosURIKey = @"op_tos_uri";
 
 - (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
   NSError *error;
-  NSDictionary *dictionary = [[NSDictionary alloc] initWithCoder:aDecoder];
+  NSDictionary *dictionary = [aDecoder decodeObjectOfClasses:[NSSet setWithObjects:[NSArray class], [NSDictionary class], nil] forKey:kDiscoveryDictionaryKey];
+
   self = [self initWithDictionary:dictionary error:&error];
   if (error) {
     return nil;
@@ -192,7 +197,7 @@ static NSString *const kOPTosURIKey = @"op_tos_uri";
 }
 
 - (void)encodeWithCoder:(NSCoder *)aCoder {
-  [_discoveryDictionary encodeWithCoder:aCoder];
+  [aCoder encodeObject:_discoveryDictionary forKey:kDiscoveryDictionaryKey];
 }
 
 #pragma mark - Properties
--- a/UnitTests/OIDServiceDiscoveryTests.m
+++ b/UnitTests/OIDServiceDiscoveryTests.m
@@ -373,13 +373,26 @@ static NSString *const kDiscoveryDocumentNullField =
 - (void)testSecureCoding {
   NSError *error;
   NSDictionary *serviceDiscoveryDictionary = [[self class] completeServiceDiscoveryDictionary];
-  OIDServiceDiscovery *discovery =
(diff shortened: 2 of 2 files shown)


WilliamDenniss:
Good catch.

Can you sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/) so we can review & merge this?

diejmon:
> Good catch.
> 
> Can you sign the [CLA](http://openid.net/intellectual-property/contribution-license-agreement/) so we can review & merge this?

Sure.

diejmon:
Will do. I will also add tests to make sure the fix works.

diejmon:
Unfortunately, I didn't found a way to accomplish what I wanted with backward compatibility. @WilliamDenniss up to you to decide what to do with this issue. For my code, I removed secure coding when using AppAuth.

WilliamDenniss:
@StevenEWright any thoughts on this?

StevenEWright:
Can we avoid catching exceptions and still maintain backwards compatibility by checking like:

```
if ([aDecoder containsValueForKey: forKey:kDiscoveryDictionaryKey]) {
    dictionary = [aDecoder decodeObjectOfClasses:[NSSet setWithObjects:[NSArray class], [NSDictionary class], nil] forKey:kDiscoveryDictionaryKey];
} else {
    dictionary = [[NSDictionary alloc] initWithCoder:aDecoder];
}
```

StevenEWright:
OK... we are talking about this offline, and I just wanted to update the thread with more info.

Here is where the problem starts: https://github.com/openid/AppAuth-iOS/blob/65ef95c6b38c7889fdda7a22f242fa0156fd43b2/Source/OIDServiceConfiguration.m#L161

This is where we decode the `OIDServiceDiscovery`. Unfortunately, the way it is currently working, the decoder is simply passed along to `NSDictionary`'s ``initWithCoder:`... but the decoder is specifically meant for an `OIDServiceDiscovery`, not an `NSDictionary`.

TL;DR: I don't believe any current serialized `OIDServiceConfiguration` will decode properly.

This may change the calculus of "backwards compatibility". If no current serialized data is decodable - then maintaining "backwards compatibility" is a moot point.

We need to understand if there are other cases where this wouldn't be a problem (like maybe persisting `OIDServiceDiscovery` instances directly?) I'm just ruminating here... I don't know... we need to look.

But, if there are no current use cases where this does actually work, then we should just fix it properly and not worry about the "backwards compatibility" issue.

Still looking....

WilliamDenniss:
Steve and I spent a bit of time debugging this. Our conclusion is that the root object of your discovery doc JSON was likely an array, and not a dictionary.

I'm remiss for not asking for that data in the first place, but we reverse-engineered your code to assume that your doc must be an array.

```
    dictionary = [aDecoder decodeObjectOfClasses:[NSSet setWithObjects:[NSArray class], [NSDictionary class], nil] forKey:kDiscoveryDictionaryKey];
```

This line means that you'll accept an `NSArray` OR an `NSDictionary`, NOT an `NSDictionary` that can contain an array.  In fact, the discovery doc can only be an `NSDictionary` and not an `NSArray` (although it can, and in fact typically will, include `NSArray`s).

The fact that your discovery doc failed to decode, but then "succeeded" with your fix, we believe indicates it was an NSArray (else why would adding `NSArray` fix it?). To catch this issue earlier (i.e. before deserialization), the `OIDServiceDiscovery` constructor now verifies that the root JSON object is an `NSDictionary` #399

The official samples serialize and deserialize `OIDServiceDiscovery` objects, and when tested against Google do not show any secure coding exceptions.

PR #318, opened 2018-10-17
Add a "Core" subspec to support Extensions

WilliamDenniss:
Adds a `AppAuth/Core` subspec that can be used by Extensions with CocoaPods
Adds a "Extensions" framework for use by Extensions building with Carthage

Resolves #198

PR #320, opened 2018-10-18
Create an AppAuth Core subspec & Framework target

WilliamDenniss:
New attempt, creates a separate include header file (AppAuthCore.h) for the subpsec and Framework targets, rather than attempting to use a single header with pre-processor defines.

This version passes `pod spec lint`, though in the 1.5.x release of CocoaPods there is an unrelated error (#7708) which is a KI. Linting with 1.6.0.beta.2 works, and the 1.5 version of CocoaPods can still be used with the library.

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -33,16 +33,26 @@ It follows the OAuth 2.0 for Native Apps best current practice
   s.platforms    = { :ios => "7.0", :osx => "10.9", :watchos => "2.0", :tvos => "9.0" }
 
   s.source       = { :git => "https://github.com/openid/AppAuth-iOS.git", :tag => s.version }
-
-  s.source_files = "Source/*.{h,m}"
   s.requires_arc = true
 
-  # iOS
-  s.ios.source_files      = "Source/iOS/**/*.{h,m}"
-  s.ios.deployment_target = "7.0"
-  s.ios.frameworks        = "SafariServices", "AuthenticationServices"
+  # Subspec for the core AppAuth library classes only, suitable for extensions.
+  s.subspec 'Core' do |core|
+     core.source_files = "Source/*.{h,m}"
+     core.exclude_files = "Source/AppAuth.h"
+  end
 
-  # macOS
-  s.osx.source_files = "Source/macOS/**/*.{h,m}"
-  s.osx.deployment_target = '10.9'
+  # Subspec for the full AppAuth library, including platform-dependant external user agents.
+  s.subspec 'ExternalUserAgent' do |externalUserAgent|
+
+    externalUserAgent.source_files = "Source/*.{h,m}"
+    
+    # iOS
+    externalUserAgent.ios.source_files      = "Source/iOS/**/*.{h,m}"
+    externalUserAgent.ios.deployment_target = "7.0"
+    externalUserAgent.ios.frameworks        = "SafariServices", "AuthenticationServices"
+
+    # macOS
+    externalUserAgent.osx.source_files = "Source/macOS/**/*.{h,m}"
+    externalUserAgent.osx.deployment_target = '10.9'    
+  end
 end
--- a/Source/AppAuthCore.h
+++ b/Source/AppAuthCore.h
@@ -1,0 +1,43 @@
+/*! @file AppAuthCore.h
+    @brief AppAuth iOS SDK
+    @copyright
+        Copyright 2015 Google Inc. All Rights Reserved.
+    @copydetails
+        Licensed under the Apache License, Version 2.0 (the "License");
+        you may not use this file except in compliance with the License.
+        You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+        Unless required by applicable law or agreed to in writing, software
+        distributed under the License is distributed on an "AS IS" BASIS,
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+ */
+
+#import "OIDAuthState.h"
+#import "OIDAuthStateChangeDelegate.h"
+#import "OIDAuthStateErrorDelegate.h"
+#import "OIDAuthorizationRequest.h"
+#import "OIDAuthorizationResponse.h"
+#import "OIDAuthorizationService.h"
+#import "OIDError.h"
+#import "OIDErrorUtilities.h"
+#import "OIDExternalUserAgent.h"
+#import "OIDExternalUserAgentRequest.h"
+#import "OIDExternalUserAgentSession.h"
+#import "OIDGrantTypes.h"
+#import "OIDIDToken.h"
+#import "OIDRegistrationRequest.h"
+#import "OIDRegistrationResponse.h"
+#import "OIDResponseTypes.h"
+#import "OIDScopes.h"
+#import "OIDScopeUtilities.h"
+#import "OIDServiceConfiguration.h"
+#import "OIDServiceDiscovery.h"
+#import "OIDTokenRequest.h"
+#import "OIDTokenResponse.h"
+#import "OIDTokenUtilities.h"
+#import "OIDURLSessionProvider.h"
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 Docs/
 xcuserdata/
+Carthage/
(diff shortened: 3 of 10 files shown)


schroepf:
With this PR I can still reproduce the same crash as in my [comment](https://github.com/openid/AppAuth-iOS/pull/318#discussion_r226406272) in #318 and my [other comment](https://github.com/openid/AppAuth-iOS/pull/310#discussion_r224952859) in #310 :

```
2018-10-18 22:07:42.719228+0200 Example-iOS_ObjC[43046:3998535] +[OIDAuthState authStateByPresentingAuthorizationRequest:presentingViewController:callback:]: unrecognized selector sent to class 0x10c1c0af8
```

I think there is an issue in cocoapods when using `use_frameworks!` and objective c categories inside a subspec. It looks like for the subspec's framework the `-ObjC` linker flag doesn't work.

WilliamDenniss:
Does the Example-iOS_ObjC sample work for you? It works for me, and doesn't use `use_frameworks!`.  The `authStateByPresentingAuthorizationRequest` method is used in the sample, FYI.

Xcode 10.0, pod 1.6.0.beta.2

schroepf:
Yes, the "vanilla" `Example-iOS_ObjC` works. It crashes only when I modify the podfile to include:
- `use_frameworks!`
- and a second target which depends on `pod 'AppAuth/Core'`

```
abstract_target 'CommonPods' do
    use_frameworks!
    
    target 'Example-iOS_ObjC' do
        platform :ios, '9.0'
        pod 'AppAuth', :path => '../../'
    end
    
    target 'Extension' do
        platform :ios, '9.0'
        pod 'AppAuth/Core', :path => '../../'
    end
end
```

Xcode 10.0, pod: 1.5.3

WilliamDenniss:
Can you contribute a test sample with a similar setup (i.e. an extension target) that we can use to test this?

julienbodet:
@schroepf I reproduced the crash you are experiencing by configuring my example the same way as you. I'll investigate to try to figure out what's happening.

WilliamDenniss:
Odd, I'm seeing the "good" crash due to the nil with that code.

Also when I follow your [steps](https://github.com/openid/AppAuth-iOS/pull/318#discussion_r226406272) to configure my sample, it doesn't crash. @julienbodet I'm curious to learn what you find.

Btw, that category adds a single, very simple method. So a workaround is not to use the category method, but rather call the code directly:
```
    OIDExternalUserAgentIOS *externalUserAgent =
        [[OIDExternalUserAgentIOS alloc]
            initWithPresentingViewController:presentingViewController];
  return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                       externalUserAgent:externalUserAgent
                                                callback:callback];
```

Still, it would be nice to fix.

schroepf:
It's weird that the crash only happens for me and @julienbodet but not for @WilliamDenniss - but at least I'm not seeing ghosts... :)

If it helps: I ran the provided sample app in the following setups:
- Xcode: 10.0
- Cocoapods: 1.5.3
- Device: iPhone Simulator(iPhone 8) running iOS 12

WilliamDenniss:
What happens with [CocoaPods 1.6](http://blog.cocoapods.org/CocoaPods-1.6.0-beta/)?

`gem install cocoapods --pre`

schroepf:
oh, yes: using cocoapods 1.6.0.beta.2 fixed the issue (for both the provided sample app as well as my actual project)

WilliamDenniss:
Mystery solved!

So it would appear this is a CocoaPods issue. FYI, I switched to the 1.6 beta to get a fix for a separate issue in the 1.5 branch with `pod spec lint`. I don't think we should workaround issues that are already fixed in CocoaPods, therefore I think we can continue. Do you agree?

WilliamDenniss:
And we still have the workaround I proposed for anyone who can't use the beta CocoaPods.

schroepf:
from the cocoapods [changelog](https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md) of 1.6.0.beta.1 (2018-08-16):

> Generate -ObjC in OTHER_LDFLAGS for apps with static frameworks.
Paul Beusterien #7946

WilliamDenniss:
Ah ha! And can you add -ObjC manually if you want to stay on Cocoapods 1.5?

schroepf:
> Ah ha! And can you add -ObjC manually if you want to stay on Cocoapods 1.5?

I tried, but it didn't work...

> So it would appear this is a CocoaPods issue. FYI, I switched to the 1.6 beta to get a fix for a separate issue in the 1.5 branch with pod spec lint. I don't think we should workaround issues that are already fixed in CocoaPods, therefore I think we can continue. Do you agree?

For me it is ok to keep this as a known issue, yes.

WilliamDenniss:
Alright. So back to the design of this – have we landed on the best option here?

I think it's neat:
* there's no additional Pod to maintain
* No pre-processor defines needed (due to the separate AppAuthCore.h)
* No changes required for anyone who doesn't care about this (they still import AppAuth.h as before)
* Works with both CocoaPods (AppAuth/Core) and Carthage (AppAuthCore framework)

Did I miss anything?

Source/AppAuthCore.h
```diff
@@ -0,0 +1,43 @@
+/*! @file AppAuth.h
```
schroepf: typo, should be `AppAuthCore.h`

WilliamDenniss: thanks, fixed.

WilliamDenniss:
@julienbodet please review if you get a chance.

julienbodet:
@WilliamDenniss Yes I’ll once back home!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

AppAuth.podspec
```diff
@@ -33,16 +33,26 @@ It follows the OAuth 2.0 for Native Apps best current practice
...
-  s.ios.deployment_target = "7.0"
-  s.ios.frameworks        = "SafariServices", "AuthenticationServices"
-
-  # macOS
-  s.osx.source_files = "Source/macOS/**/*.{h,m}"
-  s.osx.deployment_target = '10.9'
+  # Subspec for the core AppAuth library classes only, suitable for extensions.
+  s.subspec 'Core' do |core|
+     core.source_files = "Source/*.{h,m}"
+     core.exclude_files = "Source/AppAuth.h"
```
julienbodet: When installing `pod 'AppAuth'` without specifying any subspec, I understood that it takes all subspecs if there is no `default_subspec` defined.

But the tests shows that `AppAuth.h` is included, so I guess this line gets overridden by the ExternalUserAgent subspec. Seems good 🙂.

julienbodet:
Nice findings! I had to leave and didn't had time to investigate further...

@schroepf I tested again after upgrading cocoapods to the beta version and it works also for me with the example!

@WilliamDenniss I checked out your branch and noticed that AppAuthCore framework target is checked for all iOS specific files. I guess it's an error. I added another comment about something not very clear to me in the podspec. Otherwise, LGTM.

After that is merged, I think it would be nice to add extensions to the examples. For that I can probably cherry pick some commits from #224 and create a new one.

WilliamDenniss:
@julienbodet nice catch, fixed.

@julienbodet @schroepf please take a final look & LGTM if you're happy.

A sample would be awesome to have! Bonus points if it actually uses the authorization token to do something in the extension (e.g. call userinfo!).

AppAuth.podspec
WilliamDenniss: Yes, this was my theory as well. The two statements are taken together to build the list of files, so the exclusion only applies at the sub-spec level.

We could actually make ExternalUserAgent the default since it's completely stand-alone. The official docs recommend that the main spec include all the sub-specs which is why I didn't. I guess it doesn't really make much difference.

If you look at the xcodeworkspace with the sample, the ExternalUserAgent folder just contains the files that AppAuthCore doesn't, which seems to be the right outcome there at least.

schroepf:
I tested all my projects with and without extensions and they all seem to work just fine now. I also tried to setup a quick Carthage sample with an extension and AppAuthCore dependency and this project also built just fine. So yes: LGTM 👍

julienbodet:
I also tested my project with and now it builds for the extensions importing AppAuth! LGTM 🎉

@WilliamDenniss I'll make a PR for the examples next month as won't be able to do it within the next two weeks.

WilliamDenniss:
Alright, code is in. 1.0.0.beta2 :)

PR #325, opened 2018-10-26
Add option to prefer Safari View Controller

### LINKED ISSUE #209: Support sign on without SSO (no SFAuthenticationSession)
I currently use SSO with my app and SFSafariViewController webviews within my app. In order to support this you recommended hacking the code https://github.com/openid/AppAuth-iOS/issues/182 but have said you may add support in the future? If this is something you could support I'd be very happy! 

I'm currently using this library through react-native-app-auth which makes hacking the code tricky!

denisw:
### What

This PR adds a static `setPreferSafariViewController:` method to `OIDExternalUserAgentIOS`. If set to `YES`, the Safari View Controller is used even iOS 11 and 12 where `SFAuthenticationSession` and `ASWebAuthenticationSession` are normally used.

This should fix  #209.

### Why

We need to implement identity provider logout in an app. This requires us to open a special logout URL in a browser that shared the session (cookies) with the login browser's session. This is trivially possible with Safari View Controller, but not `SFAuthenticationSession` (see #209).

### DIFF
--- a/Source/iOS/OIDExternalUserAgentIOS.h
+++ b/Source/iOS/OIDExternalUserAgentIOS.h
@@ -41,6 +41,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface OIDExternalUserAgentIOS : NSObject<OIDExternalUserAgent>
 
+/*! @brief Set whether to prefer using @c SFSafariViewController even if newer user agents are available. False by default.
+    @remarks Useful if you rely on customization via @c setSafariViewControllerFactory:
+        or need to share the browser session (e.g., cookies) with other Safari View
+        Controllers.
+ */
++ (void)setPreferSafariViewController:
+    (BOOL)preferSafariViewController;
+
 /*! @brief Allows library consumers to change the @c OIDSafariViewControllerFactory used to create
         new instances of @c SFSafariViewController.
     @remarks Useful for customizing tint colors and presentation styles.
--- a/Source/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/iOS/OIDExternalUserAgentIOS.m
@@ -27,6 +27,10 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** @brief Whether to prefer @c SFSafariViewController over other user agents.
+ */
+static BOOL gPreferSafariViewController = NO;
+
 /** @brief The global/shared Safari view controller factory. Responsible for creating all new
         instances of @c SFSafariViewController.
  */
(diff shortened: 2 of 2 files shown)


WilliamDenniss:
You can do this already in AppAuth, see here: https://github.com/openid/AppAuth-iOS/issues/209#issuecomment-404353236

This class *could* be added to AppAuth, but I'm not sure it's needed.

Why can't you use logout with ASWebAuthenticationSession? I would think it should work (and have you tried the `dev-logout` branch?

denisw:
No, I didn't try the `dev-logout` branch yet - if that provides logout on all of iOS 9 - 12, then that would solve my logout problems! Thanks for the pointer.

I know that logout would work with `ASWebAuthenticationSession` (shared cookie jar with Safari), but not with `SFAuthenticationSession` (shares cookie jar only with itself in other apps). So iOS 11 is still a problem, unless I force the Safari View Controller on that version (which this PR allows).

There is another argument for adding this flag to the core codebase - the `setSafariViewControllerFactory:` API. If a developer relies on the customisations made using this method, it's kind of surprising if they just get ignored on newer iOS releases (because Safari View Controller is not used). Being able to force the Safari View Controller would be good in this case.

denisw:
So I didn't have the chance to test the `dev-logout` branch, but from the code it seems to me that it uses the same external user agent for logout as it does for login. Doesn't that mean that in case of `SFAuthenticationSession` and `ASWebAuthenticationSession`, it would still show the "[App] wants to log in with [IdP]" message although you are logging _out_? If that's the case, I and others would probably still want the option to go all-SafariViewController easily for the time being.

WilliamDenniss:
Yes, logout uses the same external user-agent by default as login. 

> the setSafariViewControllerFactory: API. If a developer relies on the customisations made using this method, it's kind of surprising if they just get ignored on newer iOS releases (because Safari View Controller is not used)

Good point that this API is confusing now. I think we should probably just remove it. Anyone who needs customizations can still achieve this by providing their own `OIDExternalUserAgent` implementation

> There is another argument for adding this flag to the core codebase 

The flag isn't needed. AppAuth supports pluggable external user-agents, you can drop your preferred one right in (including, for logout). I shared an example SFSafariViewController one that I created. If AppAuth was to include built-in SFSVC support, we could just ship that class directly. For now we don't, but I'll add your use-case to the tally, we might include it if there's enough demand & valid use-cases.

Rather than overloading the default user agent `OIDExternalUserAgentIOS` with a bunch of different modes (like SFSVC only), I think the better approach is to provide multiple implementations of `OIDExternalUserAgent`. For example, we ship with one that will use Firefox or Chrome, instead of Safari.

> still want the option to go all-SafariViewController easily for the time being

Fair enough.

Did you try the code I shared? That will allow you to do this without forking.

WilliamDenniss:
The `setSafariViewControllerFactory` API has been removed from the 1.0 branch to reduce confusion (it's only relevant for iOS 10 and below, which is increasingly not very relevant). The [gist](https://gist.github.com/WilliamDenniss/18f3779b4a310361bb955cf4e534f29c) was updated to add this API there, for those who want the SFSafariViewController user agent only.

WilliamDenniss:
Closing this PR, as the functionality is [possible](https://gist.github.com/WilliamDenniss/18f3779b4a310361bb955cf4e534f29c) without these changes needed.

PR #337, opened 2018-11-21
Examples with extensions

julienbodet:
This pull request adds extensions to all iOS examples by using the Core subspec (cocoapods) or Core framework (Carthage).

The extension is a simple UI with a logging view just like the app and allows to get the user profile.

I also discovered that the carthage example's extension is, surprisingly, able to build by importing the `AppAuth.framework` in contrary to the cocoapods example that is not building (it's the reason why the Core subspec exists). But importing `AppAuthCore.framework` instead fixes a linking warning.


petea:
@julienbodet does this work need to be updated in order to sync with head?  It'd be nice to have App Extension examples.

Toppole6:
[]()

PR #341, opened 2018-11-28
Fix incorrect error message for issuedAt validation

nickskull:
This has been changed to 10 minutes a while ago, however the error was still saying 5 minutes.


WilliamDenniss:
Closing in favor of #397

PR #345, opened 2018-12-07
Adds a completion to the session cancel. Also ensure to call completi…

### LINKED ISSUE #280: OIDExternalUserAgentIOS.dismissExternalUserAgentAnimated:completion: doesn't call completion when no current auth flow
**Are you filing an issue about iOS 12?**
No.

**Describe the bug**
`OIDExternalUserAgentIOS.dismissExternalUserAgentAnimated:completion:` doesn't execute completion block if there is no authorization flow in progress (`_externalUserAgentFlowInProgress` is `false`). I believe the completion block should be called in every case and could maybe return a boolean to tell the caller if the authorization agent was actually dismissed or not (`true` if a flow was in progress, `false` otherwise).

```
- (void)dismissExternalUserAgentAnimated:(BOOL)animated completion:(void (^)(void))completion {
  if (!_externalUserAgentFlowInProgress) {
    // Ignore this call if there is no authorization flow in progress.
    return;
  }
  
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wpartial-availability"
  SFSafariViewController *safariVC = _safariVC;
  SFAuthenticationSession *authenticationVC = _authenticationVC;
#pragma clang diagnostic pop
  
  [self cleanUp];
  
  if (@available(iOS 11.0, *)) {
    [authenticationVC cancel];
    if (completion) completion();
  } else if (@available(iOS 9.0, *)) {
    if (safariVC) {
      [safariVC dismissViewControllerAnimated:YES completion:completion];
    } else {
      if (completion) completion();
    }
  } else {
    if (completion) completion();
  }
}
```

I noticed this completion block was not called while calling `dismissExternalUserAgentAnimated:completion:` because I wanted to cancel the authorization in a certain case and execute some code after it is actually cancelled or not (if no flow in progress). The preferred way to do would be to call `OIDExternalUserAgentSession.cancel()` but this method does not currently allow to pass a completion block to execute once the authorization is cancelled. That said, adding a completion to this cancel method would be nice too.

julienbodet:
…on block in all the implementations of dismissExternalUserAgentAnimated:completion:

This pull request resolves #280

### DIFF
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -79,12 +79,13 @@ NS_ASSUME_NONNULL_BEGIN
   }
 }
 
-- (void)cancel {
+- (void)cancelWithCompletion:(void (^)(void))completion {
   [_externalUserAgent dismissExternalUserAgentAnimated:YES completion:^{
       NSError *error = [OIDErrorUtilities errorWithCode:OIDErrorCodeUserCanceledAuthorizationFlow
                                         underlyingError:nil
                                             description:@"Authorization flow was cancelled."];
       [self didFinishWithResponse:nil error:error];
+      if (completion) completion();
   }];
 }
 
--- a/Source/OIDAuthorizationService.m
+++ b/Source/OIDAuthorizationService.m
@@ -79,7 +79,11 @@ NS_ASSUME_NONNULL_BEGIN
   }
 }
 
-- (void)cancelWithCompletion:(void (^)(void))completion {
+- (void)cancel {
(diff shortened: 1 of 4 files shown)


WilliamDenniss:
Should `completion` be nullable, and we keep the existing "cancel" method to avoid a breaking API change?

julienbodet:
> Should `completion` be nullable, and we keep the existing "cancel" method to avoid a breaking API change?

Yes good idea, I'll make the change!

PR #354, opened 2019-02-22
Clean up test scopes

petea:


WilliamDenniss:
The "openid" scope has special meaning, I wonder if for clarity we should pick something else, or use it without the prefix.

petea:
How about "userinfo.email"?

WilliamDenniss:
That works.

petea:
Ready for you to take another look.

PR #368, opened 2019-03-15
Fixes issue with Guided Access or Single App modes  (rdar://40809553)

### LINKED ISSUE #365: Log in doesn't work in Guided Access or Single App modes
**Describe the bug**
Log in doesn't work in Guided Access mode or single app mode with supervised devices under an MDM. As per https://openradar.appspot.com/48376122, both ASWebAuthenticationSession and SFAuthenticationSession return an error (Domain=com.apple.AuthenticationServices.WebAuthenticationSession Code=1 "(null)") in these cases.

**To Reproduce**
1. Enable guided access on a device (Settings -> General -> Accessibility)
2. Set up and run the example project.
3. Start guided access mode on device (Tap home/power button 3 times, depending on device type)
4. Tap button "Authorize" in the app.

**Expected behavior**
Have an alert window to continue sign-in, and then the webview should appear.

** Actual behavior**
Neither the alert window nor the webview are presented, and an error occurs (Domain=com.apple.AuthenticationServices.WebAuthenticationSession Code=1 "(null)").

**Smartphone:**
 - Device: iPhone8
 - OS: iOS 12.1.4
 - XCode: 10.1 10B61

Tested with library versions 0.95.0 and master.

cassianodialpad:
Fixes #365

### DIFF
--- a/Source/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/iOS/OIDExternalUserAgentIOS.m
@@ -61,7 +61,7 @@ NS_ASSUME_NONNULL_BEGIN
 
   _externalUserAgentFlowInProgress = YES;
   _session = session;
-  BOOL openedSafari = NO;
+  BOOL openedUserAgent = NO;
   NSURL *requestURL = [request externalUserAgentRequestURL];
 
   // iOS 12 and later, use ASWebAuthenticationSession
@@ -91,13 +91,13 @@ NS_ASSUME_NONNULL_BEGIN
         }
       }];
       _webAuthenticationVC = authenticationVC;
-      openedSafari = [authenticationVC start];
+      openedUserAgent = [authenticationVC start];
     }
   }
   // iOS 11, use SFAuthenticationSession
   // SFAuthenticationSession doesn't work with guided access (rdar://40809553)
   if (@available(iOS 11.0, *)) {
-    if (!openedSafari && !UIAccessibilityIsGuidedAccessEnabled()) {
+    if (!openedUserAgent && !UIAccessibilityIsGuidedAccessEnabled()) {
       __weak OIDExternalUserAgentIOS *weakSelf = self;
       NSString *redirectScheme = request.redirectScheme;
       SFAuthenticationSession *authenticationVC =
@@ -121,32 +121,33 @@ NS_ASSUME_NONNULL_BEGIN
         }
       }];
       _authenticationVC = authenticationVC;
-      openedSafari = [authenticationVC start];
+      openedUserAgent = [authenticationVC start];
     }
   }
   // iOS 9 and 10, use SFSafariViewController
   if (@available(iOS 9.0, *)) {
-    if (!openedSafari) {
+    if (!openedUserAgent) {
       SFSafariViewController *safariVC =
           [[SFSafariViewController alloc] initWithURL:requestURL];
       safariVC.delegate = self;
       _safariVC = safariVC;
       [_presentingViewController presentViewController:safariVC animated:YES completion:nil];
-      openedSafari = YES;
+      openedUserAgent = YES;
     }
+  }
   // iOS 8 and earlier, use mobile Safari
-  } else {
-    openedSafari = [[UIApplication sharedApplication] openURL:requestURL];
+  if (!openedUserAgent){
+    openedUserAgent = [[UIApplication sharedApplication] openURL:requestURL];
   }
 
-  if (!openedSafari) {
+  if (!openedUserAgent) {
     [self cleanUp];
     NSError *safariError = [OIDErrorUtilities errorWithCode:OIDErrorCodeSafariOpenError
                                             underlyingError:nil
                                                 description:@"Unable to open Safari."];
     [session failExternalUserAgentFlowWithError:safariError];
   }
-  return openedSafari;
+  return openedUserAgent;
 }
 
 - (void)dismissExternalUserAgentAnimated:(BOOL)animated completion:(void (^)(void))completion {
--- a/Source/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/iOS/OIDExternalUserAgentIOS.m
@@ -64,9 +64,9 @@ NS_ASSUME_NONNULL_BEGIN
   BOOL openedUserAgent = NO;
   NSURL *requestURL = [request externalUserAgentRequestURL];
 
-  // iOS 12 and later, use ASWebAuthenticationSession
-  // ASWebAuthenticationSession doesn't work with guided access (rdar://40809553)
   if (@available(iOS 12.0, *)) {
+    // iOS 12 and later, use ASWebAuthenticationSession
+    // ASWebAuthenticationSession doesn't work with guided access (rdar://40809553)
     if (!UIAccessibilityIsGuidedAccessEnabled()) {
       __weak OIDExternalUserAgentIOS *weakSelf = self;
       NSString *redirectScheme = request.redirectScheme;
@@ -94,9 +94,9 @@ NS_ASSUME_NONNULL_BEGIN
       openedUserAgent = [authenticationVC start];
     }
   }
-  // iOS 11, use SFAuthenticationSession
-  // SFAuthenticationSession doesn't work with guided access (rdar://40809553)
   if (@available(iOS 11.0, *)) {
+    // iOS 11, use SFAuthenticationSession
+    // SFAuthenticationSession doesn't work with guided access (rdar://40809553)
     if (!openedUserAgent && !UIAccessibilityIsGuidedAccessEnabled()) {
       __weak OIDExternalUserAgentIOS *weakSelf = self;
       NSString *redirectScheme = request.redirectScheme;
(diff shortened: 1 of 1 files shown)


WilliamDenniss:
I gather this fixes the issue by opening Safari?

Can you sign the [CLA ](https://openid.net/intellectual-property/contribution-license-agreement/)or link me to your one if you have already?

cassianodialpad:
Yes, it will open Safari when in Guided Access.

My company has sent the corporate CLA to help@oidf.org , but it seems it has not been processed yet (Dialpad is not on the list).

cassianodialpad:
Hi @WilliamDenniss , here is the CLA: https://openid.net/wordpress-content/uploads/2019/03/Dialpad-Inc_CLA-Agreement-Corporate_2019-03-18.pdf

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -65,65 +65,75 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
   NSURL *requestURL = [request externalUserAgentRequestURL];
 
   // iOS 12 and later, use ASWebAuthenticationSession
+  // ASWebAuthenticationSession doesn't work with guided access (rdar://40809553)
   if (@available(iOS 12.0, *)) {
```
WilliamDenniss: I would just add the extra condition here, i.e. `&& !UIAccessibilityIsGuidedAccessEnabled()` makes for less indentation.

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -65,65 +65,75 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
-                             underlyingError:error
-                                 description:@"User cancelled."];
-        [strongSelf->_session failExternalUserAgentFlowWithError:safariError];
-      }
-    }];
-    _authenticationVC = authenticationVC;
-    openedSafari = [authenticationVC start];
+  // SFAuthenticationSession doesn't work with guided access (rdar://40809553)
+  if (@available(iOS 11.0, *)) {
+    if (!openedSafari && !UIAccessibilityIsGuidedAccessEnabled()) {
```
WilliamDenniss: same comment as above for `&& !UIAccessibilityIsGuidedAccessEnabled()`.

WilliamDenniss: I don't see how `openedSafari` can ever be true at this point is that check needed?

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -154,21 +164,17 @@ - (void)dismissExternalUserAgentAnimated:(BOOL)animated completion:(void (^)(voi
   
   [self cleanUp];
   
-  if (@available(iOS 12.0, *)) {
```
WilliamDenniss: I like the refactor of this whole section here.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Actually Xcode gives a compilation error when chaining conditions with `@available`, see https://stackoverflow.com/questions/46965347/objective-c-available-guard-anded-with-more-conditions.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Same.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Thanks. It actually wasn't just a refactor, but to prevent problems if the user switch on/off guided access in the middle of the login process.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: True, I will fix that.

Source/iOS/OIDExternalUserAgentIOS.m
WilliamDenniss: Oh, not saying it wasn't needed - I just like this improvement even without the guided access fixes, I think it's the better way to structure it generally.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Actually it is needed in this context indeed: if on iOS12 not in access mode, `openedSafari` will have the result of `[authenticationVC start]`. If it did not work with ASWebAuthenticationSession for some reason, it will still try SFAuthenticationSession.
What do you think? Is there a better way to handle this?

WilliamDenniss: Ah ok, I missed that the `else if` was removed.

The new logic is a little harder to read, but it's also inconsistent. In most places "openedSafari" is the test rather than the platform version, but in the case of mobile Safari there's no `openedSafari` check, and happens only for iOS 8 & below like before (mobile safari is opened in the `else` path of the iOS 9 version check).

I think we need to holistically look at the logic here and move to consistent model, not just apply a fix for UIAccessibilityIsGuidedAccessEnabled.

Perhaps that consistent model is "did the previous preferred method work, otherwise try the next best option", which is almost what you've done here, with the exception of mobile Safari.

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -65,65 +65,75 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
+        } else {
+          NSError *safariError =
+              [OIDErrorUtilities errorWithCode:OIDErrorCodeUserCanceledAuthorizationFlow
+                               underlyingError:error
+                                   description:nil];
+          [strongSelf->_session failExternalUserAgentFlowWithError:safariError];
+        }
+      }];
+      _webAuthenticationVC = authenticationVC;
+      openedSafari = [authenticationVC start];
```
WilliamDenniss: Can you rename `openedSafari` to `openedUserAgent` for accuracy?

cassianodialpad: Sure!

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Makes sense. I didn't want to modify the code too much for that fix, but you are right. I will make the changes.

cassianodialpad: I added the condition on the `else` path. Do you think we need a larger refactor to this model?

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -61,82 +61,93 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
 
   _externalUserAgentFlowInProgress = YES;
   _session = session;
-  BOOL openedSafari = NO;
+  BOOL openedUserAgent = NO;
   NSURL *requestURL = [request externalUserAgentRequestURL];
 
   // iOS 12 and later, use ASWebAuthenticationSession
+  // ASWebAuthenticationSession doesn't work with guided access (rdar://40809553)
```
WilliamDenniss: Can you move this comment to just above the if-statement for `UIAccessibilityIsGuidedAccessEnabled`? And below

Source/iOS/OIDExternalUserAgentIOS.m
WilliamDenniss: I think it's good to go. The approach seems valid, it's basically a cascading order of preferences that's less attached to the iOS version now (while still gated by the version).

WilliamDenniss: Can you address the nit regarding the comment, and squish the commits into 1?

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: Sure.

Source/iOS/OIDExternalUserAgentIOS.m
cassianodialpad: You don't squash for merging pull requests?

Source/iOS/OIDExternalUserAgentIOS.m
WilliamDenniss: Sorry I just meant just the UIAccessibilityIsGuidedAccessEnabled comment, not both. I think the comment should precede the if-statement that it relates to.

cassianodialpad: Done! =)

cassianodialpad:
Thanks!
How do I make a PR for a fix to the current version?

WilliamDenniss:
Assuming no commits: stash any changes, checkout HEAD, pop the stash and continue.

cassianodialpad:
Sorry, I did't get it... The current version is 0.95.0, is it possible to create a 0.95.1 with this fix? From what I understand from github, I can't make a PR to other branch than master (or to create a new one).

WilliamDenniss:
I thought you were referring to further changes on this branch.

0.95.0 is a tag, we release it from the [pre1.0 ](https://github.com/openid/AppAuth-iOS/tree/pre1.0) branch which you can file a PR against, and we can cut a 0.95.1 from that.

Personally wouldn't bother backporting the fix though, anyone who needs this should just use the 1.x builds. We'll be cutting a release candidate from master soon.

cassianodialpad:
I'm using https://github.com/google/GTMAppAuth, with depends on 0.95.0 and breaks with 1.0... =/

Issue #374, opened 2019-03-24
JSON error parsing token response


mrmuhammadali:
**Describe the bug**
When using AppAuth for Github authentication on iOS, it gives me following error:
`JSON error parsing token response: The data couldn’t be read because it isn’t in the correct format.`
even though I've completed the authentication flow of Github.

WilliamDenniss:
See comment on #375, #206

PR #375, opened 2019-03-24
Fix issue #374

mrmuhammadali:
I sent accept header, i.e. Accept: application/json, in token request and it fixed the mentioned issue. Some Token Providers, i.e. Github, require this header in order to send response in JSON format.


WilliamDenniss:
Hi, thanks for the PR but unfortunately it's our policy not to accept vendor-specific changes.

Github apparently does not have a compliant OAuth 2.0 endpoint, as the response type for OAuth is always JSON. See Section 5.1 of RFC 6749:

> The parameters are included in the entity-body of the HTTP response
   using the "application/json" media type as defined by [RFC4627]

There is a longer discussion in #206 about this topic.

However, this fix is exactly what you can use locally to integrate with GitHub.

WilliamDenniss:
See comment on #375, #206

PR #380, opened 2019-03-29
Support RFC 8414 better by allowing for its OPTIONAL metadata

WilliamDenniss:


bendoppler:
hi @WilliamDenniss , do u have any estimated deadline for the time the fix can merge and release version of this fix? Thanks

yannickl:
@WilliamDenniss bumping the issue again. Cloudflare does not return all the metadata by default and it would make things much easier if it could be supported natively at the AppAuth level.

w-goog:
Thank you for the bump, Yannick! Ambiently taking a look; I think we can get this addressed after the next release.

PR #428, opened 2019-08-15
Replaced UIApplication.keyWindow usage with searching for a fitting WindowScene

KlemensStrasser:
UIApplication.keyWindow was deprecated in iOS13 since there is no single keyWindow anymore when dealing with mutliple windows. Using keyWindow can result in the authenticationVC beign presented on a window that is not currently in the foreground. So the way to solve it is to look through all connected scenes and use a window of the most fitting windowScene.

Note:
When having multiple scenes of the same app in the foreground via splitView or slideover, this solution might result in the authenticationVC beign displayed on any of these scenes. But there is currently no better solution since there is no way to tell in which scene the user has triggered the authentication.


WilliamDenniss:
Thank you for the contribution!

I think the better approach is to have the user supply the UIViewController whose's owning Window should be used as the presenting anchor. That way we let the user determine where to present it. Also, the library requires a presenting UIViewController on iOS 9 and 10 anyway.

WilliamDenniss:
Closing in favor of #429.

Thanks for this work, I didn't realize that keyWindow was deprecated, good to get this fixed up before we merge the iOS 13 support.

PR #430, opened 2019-08-16
Fix logout not working on iOS 11 and above

tonyphn:
Issue: AppAuth is using ASWebAuthenticationSession on iOS 12 and SFAuthenticationSession on iOS 11 to login, however when triggering logout using the same authentication session it doesn't clear the cached credential. 
Solution: Use SFSafariViewController on iOS 9 and above to guarantee logout always work


WilliamDenniss:
No, this is not the right approach here.

AppAuth supports pluggable UserAgents! If you need a UserAgent that behaves differently, you can simply add one. The UserAgent that you are altering is designed to support SSO.

Issue #408, opened 2019-06-21
UIKit for Mac (Catalyst) support


jeroenleenarts:
**Are you filing an issue about iOS 12?**
No

**Is your feature request related to a problem? Please describe.**
When compiling against UIKit for Mac (Catalyst, Xcode 11) the build fails with a linker error:

```
Undefined symbols for architecture x86_64:
  _OBJC_CLASS_$_SFAuthenticationSession, referenced from:
      objc-class-ref in OIDExternalUserAgentIOS.o
ld: symbol(s) not found for architecture x86_64
```

**Describe the solution you would like**
Full support for UIKit for Mac.

**Describe alternatives you have considered**
Unsure about possibilities here.

**Additional context**
.

jeroenleenarts:
Most likely it is a matter for working around the fact that SFAuthenticationSession is unavailable on UIKit for Mac.

`ASWebAuthenticationSession` is supported. So I hope/expect it is a matter of conditional compilation specific for UIKit for Mac. No other symbols used in AppAuth-iOS seem to be unsupported on UIKit for Mac.

WilliamDenniss:
I suspect we'll need a UIKit for Mac specific `OIDExternalUserAgent` implementation. Should be pretty straight forward. May need a separate target as well.

below:
While I have not added my own ClientID, I did compile and run master on macOS Catalina. Am I missing steps to reproduce the issue?

jeroenleenarts:
@below did you enable UIKIT for Mac?

jeroenleenarts:
With these steps I run into the mentioned compile error:

- clone repo
- cd into checkout location and example project `Example_Objc`
- run pod install 
- open `xcworkspace` with xcode beta
- enable mac target on target `Example_Objc`
- Select a valid Team for signing
- build

below:
> @below did you enable UIKIT for Mac?

No, how would I do that other than selecting the Mac Target for Example_Objc?

jeroenleenarts:
Xcode 11 Beta 2. Ars has a nice overview: https://arstechnica.com/gadgets/2019/07/catalyst-deep-dive-the-future-of-mac-software-according-to-apple-and-devs/

Also, this allows compiling, the example with UIKitForMac:
```
diff --git a/Source/iOS/OIDExternalUserAgentIOS.m b/Source/iOS/OIDExternalUserAgentIOS.m
index 6303d6d..593da11 100644
[...]

PR #455, opened 2019-10-15
Support for mac catalyst

julienbodet:
This PR adds support for mac catalyst (UIKit for Mac) (#408). The main changes are:
- Enabled the Mac checkbox on iOS framework targets (AppAuth_iOS and AppAuthCore) to make them compatible with mac catalyst.
- Added a new external user agent implementation for Catalyst (`OIDExternalUserAgentCatalyst`) which basically does the same thing as `OIDExternalUserAgentIOS` for iOS13. I could've added the support for Catalyst in that existing class, but it would have added too much complexity to the code.
- In the 'Compile Sources' build phase, I restricted `OIDExternalUserAgentCatalyst` to macOS and `OIDExternalUserAgentIOS`, `OIDExternalUserAgentIOSCustomBrowser` to iOS only as these two classes don't compile on macOS Catalyst.

At that point the frameworks are able to build. 

Then I needed to find a way to tell Cocoapods which user agents classes are for iOS and macOS. But the thing is that there is no new platform like 'iosmac' or 'maccatalyst' to specify in the podspec. So I did like [Apple mentions](https://developer.apple.com/documentation/xcode/creating_a_mac_version_of_your_ipad_app?language=objc) and enclosed the code using `TARGET_OS_MACCATALYST`.

I also tested using Carthage but it doesn't work. It seems that `carthage bootstrap` doesn't compile the frameworks correctly for Catalyst. Support for Catalyst is in their [roadmap](https://github.com/Carthage/Carthage/issues/2890).

### DIFF
--- a/Source/iOS/OIDAuthState+IOS.m
+++ b/Source/iOS/OIDAuthState+IOS.m
@@ -18,7 +18,13 @@
 
 #import "OIDAuthState+IOS.h"
 
-#import "OIDExternalUserAgentIOS.h"
+#if TARGET_OS_MACCATALYST
+  #import "OIDExternalUserAgentCatalyst.h"
+  typedef OIDExternalUserAgentCatalyst ExternalUserAgent;
+#else
+  #import "OIDExternalUserAgentIOS.h"
+  typedef OIDExternalUserAgentIOS ExternalUserAgent;
+#endif
 
 @implementation OIDAuthState (IOS)
 
@@ -26,9 +32,9 @@
     authStateByPresentingAuthorizationRequest:(OIDAuthorizationRequest *)authorizationRequest
                      presentingViewController:(UIViewController *)presentingViewController
                                      callback:(OIDAuthStateAuthorizationCallback)callback {
-    OIDExternalUserAgentIOS *externalUserAgent =
-        [[OIDExternalUserAgentIOS alloc]
-            initWithPresentingViewController:presentingViewController];
+  ExternalUserAgent *externalUserAgent =
+      [[ExternalUserAgent alloc]
+          initWithPresentingViewController:presentingViewController];
   return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                        externalUserAgent:externalUserAgent
                                                 callback:callback];
@@ -37,7 +43,7 @@
 + (id<OIDExternalUserAgentSession>)
     authStateByPresentingAuthorizationRequest:(OIDAuthorizationRequest *)authorizationRequest
                                   callback:(OIDAuthStateAuthorizationCallback)callback {
-  OIDExternalUserAgentIOS *externalUserAgent = [[OIDExternalUserAgentIOS alloc] init];
+  ExternalUserAgent *externalUserAgent = [[ExternalUserAgent alloc] init];
   return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                        externalUserAgent:externalUserAgent
                                                 callback:callback];
--- a/Source/iOS/OIDAuthState+IOS.m
+++ b/Source/iOS/OIDAuthState+IOS.m
@@ -17,14 +17,8 @@
  */
 
 #import "OIDAuthState+IOS.h"
-
-#if TARGET_OS_MACCATALYST
-  #import "OIDExternalUserAgentCatalyst.h"
-  typedef OIDExternalUserAgentCatalyst ExternalUserAgent;
-#else
-  #import "OIDExternalUserAgentIOS.h"
-  typedef OIDExternalUserAgentIOS ExternalUserAgent;
-#endif
+#import "OIDExternalUserAgentIOS.h"
+#import "OIDExternalUserAgentCatalyst.h"
 
 @implementation OIDAuthState (IOS)
 
@@ -32,9 +26,13 @@
     authStateByPresentingAuthorizationRequest:(OIDAuthorizationRequest *)authorizationRequest
                      presentingViewController:(UIViewController *)presentingViewController
                                      callback:(OIDAuthStateAuthorizationCallback)callback {
-  ExternalUserAgent *externalUserAgent =
-      [[ExternalUserAgent alloc]
-          initWithPresentingViewController:presentingViewController];
+  id<OIDExternalUserAgent> externalUserAgent;
+#if TARGET_OS_MACCATALYST
+  externalUserAgent = [[OIDExternalUserAgentCatalyst alloc]
+      initWithPresentingViewController:presentingViewController];
+#else
+  externalUserAgent = [[OIDExternalUserAgentIOS alloc] initWithPresentingViewController:presentingViewController];
+#endif
   return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                        externalUserAgent:externalUserAgent
                                                 callback:callback];
@@ -43,7 +41,12 @@
 + (id<OIDExternalUserAgentSession>)
     authStateByPresentingAuthorizationRequest:(OIDAuthorizationRequest *)authorizationRequest
                                   callback:(OIDAuthStateAuthorizationCallback)callback {
-  ExternalUserAgent *externalUserAgent = [[ExternalUserAgent alloc] init];
+  id<OIDExternalUserAgent> externalUserAgent;
+#if TARGET_OS_MACCATALYST
+  externalUserAgent = [[OIDExternalUserAgentCatalyst alloc] init];
+#else
+  externalUserAgent = [[OIDExternalUserAgentIOS alloc] init];
+#endif
   return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                        externalUserAgent:externalUserAgent
                                                 callback:callback];
--- a/Source/iOS/OIDAuthState+IOS.m
+++ b/Source/iOS/OIDAuthState+IOS.m
@@ -38,18 +38,15 @@
                                                 callback:callback];
 }
 
+#if !TARGET_OS_MACCATALYST
 + (id<OIDExternalUserAgentSession>)
     authStateByPresentingAuthorizationRequest:(OIDAuthorizationRequest *)authorizationRequest
                                   callback:(OIDAuthStateAuthorizationCallback)callback {
-  id<OIDExternalUserAgent> externalUserAgent;
-#if TARGET_OS_MACCATALYST
-  externalUserAgent = [[OIDExternalUserAgentCatalyst alloc] init];
-#else
-  externalUserAgent = [[OIDExternalUserAgentIOS alloc] init];
-#endif
+  OIDExternalUserAgentIOS *externalUserAgent = [[OIDExternalUserAgentIOS alloc] init];
   return [self authStateByPresentingAuthorizationRequest:authorizationRequest
                                        externalUserAgent:externalUserAgent
                                                 callback:callback];
 }
+#endif
 
 @end
--- a/Source/iOS/OIDAuthorizationService+IOS.m
+++ b/Source/iOS/OIDAuthorizationService+IOS.m
@@ -18,7 +18,13 @@
 
 #import "OIDAuthorizationService+IOS.h"
 
-#import "OIDExternalUserAgentIOS.h"
+#if TARGET_OS_MACCATALYST
+  #import "OIDExternalUserAgentCatalyst.h"
+  typedef OIDExternalUserAgentCatalyst ExternalUserAgent;
+#else
+  #import "OIDExternalUserAgentIOS.h"
+  typedef OIDExternalUserAgentIOS ExternalUserAgent;
+#endif
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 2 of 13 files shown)


Source/iOS/OIDExternalUserAgentCatalyst.m
```diff
@@ -0,0 +1,148 @@
...
+      
+  authenticationVC.presentationContextProvider = self;
+  _webAuthenticationVC = authenticationVC;
+  openedUserAgent = [authenticationVC start];
+
+  if (!openedUserAgent) {
+    [self cleanUp];
+    NSError *safariError = [OIDErrorUtilities errorWithCode:OIDErrorCodeSafariOpenError
+                                            underlyingError:nil
+                                                description:@"Unable to open Safari."];
```
WilliamDenniss: May as well be explicit in the error message since there's only one, i.e. "ASWebAuthenticationSession"

Source/iOS/OIDAuthState+IOS.m
```diff
@@ -18,17 +18,23 @@
 
 #import "OIDAuthState+IOS.h"
 
-#import "OIDExternalUserAgentIOS.h"
+#if TARGET_OS_MACCATALYST
```
WilliamDenniss: I need to learn more about Mac Catalyst. Instead of OIDAuthState+IOS supporting both iOS and Catalyst, can we instead have a OIDAuthState+MacCatalyst  that the Catalyst target imports exclusively?  Or do both build from the same target?

WilliamDenniss:
Travis is [failing](https://travis-ci.org/openid/AppAuth-iOS/jobs/598384816) with:
```
Compiling OIDRPProfileCode.m
❌  error: umbrella header for module 'AppAuth' does not include header '/Users/travis/build/openid/AppAuth-iOS/Source/iOS/OIDExternalUserAgentCatalyst.h' [-Werror,-Wincomplete-umbrella]
❌  /Users/travis/build/openid/AppAuth-iOS/UnitTests/OIDRPProfileCode.h:21:9: could not build module 'AppAuth'
```

Source/iOS/OIDExternalUserAgentCatalyst.h
```diff
@@ -0,0 +1,48 @@
+/*! @file OIDExternalUserAgentCatalyst.h
+   @brief AppAuth iOS SDK
+   @copyright
+       Copyright 2016 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
julienbodet: Fixed in new files.

Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -209,6 +211,7 @@ - (void)cleanUp {
   // them while not in an authorization flow.
   _safariVC = nil;
   _authenticationVC = nil;
+  _webAuthenticationVC = nil;
```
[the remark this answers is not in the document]
julienbodet: Reverted.

Source/iOS/OIDExternalUserAgentCatalyst.m
julienbodet: Done.

Source/iOS/OIDAuthState+IOS.m
julienbodet: Catalyst builds from the same target as the iOS framework. I don't think we can have a separate target for Catalyst (maybe I'm wrong?) or at least I don't see how that's possible.

julienbodet:
> Travis is [failing](https://travis-ci.org/openid/AppAuth-iOS/jobs/598384816) with:
> 
> ```
> Compiling OIDRPProfileCode.m
> ❌  error: umbrella header for module 'AppAuth' does not include header '/Users/travis/build/openid/AppAuth-iOS/Source/iOS/OIDExternalUserAgentCatalyst.h' [-Werror,-Wincomplete-umbrella]
> ❌  /Users/travis/build/openid/AppAuth-iOS/UnitTests/OIDRPProfileCode.h:21:9: could not build module 'AppAuth'
> ```

This happens only when running the tests. That's because `OIDExternalUserAgentCatalyst.h` is not imported for `TARGET_OS_IOS=1` in the umbrella header. The thing I don't understand is that I get this error even though I declared this header file as being supported only for macOS in the build phases of the `AppAuth_iOS` target. And in the generated framework folder I clearly see that this header is absent. When I use the generated framework by linking it in an example, everything works fine.

One way to fix this would be to import all the user agents headers no matter the platform (iOS or macCatalyst) and this fix the issue.

I'll continue digging on that, if you have any idea, let me know. Thanks :)

Source/iOS/OIDAuthorizationService+IOS.m
```diff
@@ -18,7 +18,13 @@
 
 #import "OIDAuthorizationService+IOS.h"
 
-#import "OIDExternalUserAgentIOS.h"
+#if TARGET_OS_MACCATALYST
+  #import "OIDExternalUserAgentCatalyst.h"
+  typedef OIDExternalUserAgentCatalyst ExternalUserAgent;
```
WilliamDenniss: The `typedef` pattern doesn't look right to me.

I have another suggestion for the changes to this file, and to `OIDAuthState+IOS.m`:

1. Import *both* user agents. They are #ifdef protected anyway so there's no harm in importing them.
2. Use an `#if TARGET_OS_MACCATALYST` to determine which external agent is created.

My preference for this approach is that the Catalyst-specific tweaks are localized to just the specific area of code where they matter.

Source/iOS/OIDAuthorizationService+IOS.m
```diff
@@ -27,7 +33,7 @@ @implementation OIDAuthorizationService (IOS)
 + (id<OIDExternalUserAgentSession>) presentAuthorizationRequest:(OIDAuthorizationRequest *)request
     presentingViewController:(UIViewController *)presentingViewController
                     callback:(OIDAuthorizationCallback)callback {
-  OIDExternalUserAgentIOS *externalUserAgent = [[OIDExternalUserAgentIOS alloc]
```
WilliamDenniss: Per above, I'd change this to:

```
id<OIDExternalUserAgent> externalUserAgent;
#if TARGET_OS_MACCATALYST
externalUserAgent = [[OIDExternalUserAgentCatalyst alloc] .....
```

Source/iOS/OIDExternalUserAgentCatalyst.h
```diff
@@ -0,0 +1,48 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/*! @brief A Catalyst specific external user-agent that uses `ASWebAuthenticationSession` to
+       present the request.
+*/
+@interface OIDExternalUserAgentCatalyst : NSObject<OIDExternalUserAgent>
+
+- (nullable instancetype)init API_AVAILABLE(ios(11))
+__deprecated_msg("This method will not work on iOS 13, use "
```
WilliamDenniss: Do we need this deprecated method, since this class is new?

Source/iOS/OIDAuthState+IOS.m
WilliamDenniss: Understood, thanks for clarifying.

Source/iOS/OIDExternalUserAgentIOSCustomBrowser.h
```diff
@@ -20,6 +20,8 @@
 
 #import "OIDExternalUserAgent.h"
 
+#if !TARGET_OS_MACCATALYST
```
WilliamDenniss: For discussion (am not asking for changes here):

One implication of this, is that anyone who uses OIDExternalUserAgentIOSCustomerBrowser will get missing symbols when they build for Catalyst.  Maybe that's OK, since they can't use it in any case.

Another approach would be to implement stub methods for Catalyst that just assert.

The benefit of the assert is that it can explain what to do ("you must use Catalyst specific user-agents on Catalyst"). The benefit of the missing symbols is that it's compile time rather than runtime, although the compiler message may confuse.

Source/iOS/OIDAuthorizationService+IOS.m
julienbodet: Thanks for the suggestion! I initially did that for `OIDAuthState+IOS.m` to avoid adding `#if TARGET_OS_MACCATALYST` at multiple places.

> My preference for this approach is that the Catalyst-specific tweaks are localized to just the specific area of code where they matter.
I agree, we see better now where the catalyst platform specific code is.

Source/iOS/OIDExternalUserAgentCatalyst.h
julienbodet: In `OIDAuthState+IOS.m` we have a method which doesn't require a presenting view controller, that is the reason why I kept that init until we remove the ability to not provide a view controller.

But, since this class is new, I can remove this deprecated method and assert in `OIDAuthState+IOS.m` instead for Catalyst, what I prefer.

Source/iOS/OIDExternalUserAgentIOSCustomBrowser.h
julienbodet: Personally I prefer getting an error at compile time rather than runtime, but the assert approach has a valid point for the clarity of the message. Although the current approach doesn't give a clear message to the user, the user agents names give a good indication and suggest that some are for iOS and other for Catalyst, having that in their class name.

I'm not sure what's the best practice for this use case... but I'm open to implement what is best for the user.

Source/iOS/OIDExternalUserAgentCatalyst.h
WilliamDenniss: Sounds good.

I'd probably go with the `#if !TARGET_OS_MACCATALYST`. Since the Catalyst AppAuth API is "new" we're not breaking any backwards comparability of the existing API surface by removing it. Either way is reasonable.

ykphuah:
I am eager to test this PR, is this the way to use ?

   ```
#if TARGET_OS_MACCATALYST
            let ua = OIDExternalUserAgentCatalyst()
#else
            let ua = OIDExternalUserAgentIOS(presenting: self)!
#endif
```

I get Use of unresolved identifier 'OIDExternalUserAgentIOS' when building for catalyst, building for iOS worked fine.

**Comment by julienbodet** [comment]
> ```
> #if TARGET_OS_MACCATALYST
>          let ua = OIDExternalUserAgentCatalyst()
> #else
>          let ua = OIDExternalUserAgentIOS(presenting: self)!
> #endif
> ```
> 
> I get Use of unresolved identifier 'OIDExternalUserAgentIOS' when building for catalyst, building for iOS worked fine.

`OIDExternalUserAgentIOS` is the default iOS implementation whereas `OIDExternalUserAgentCatalyst` is the default Catalyst implementation. So the error you got is normal. 

Using `TARGET_OS_MACCATALYST` is the way to go if you need to pass specific user agents implementations. But, in your case none of this is necessary, I recommend you to use [`authStateByPresentingAuthorizationRequest:presentingViewController:callback:`](https://github.com/openid/AppAuth-iOS/blob/e07ba3325e2b5e9cca1c7a022715fe98368dd18d/Source/iOS/OIDAuthState%2BIOS.h#L44) convenience method handling all of that for you and returning an `OIDAuthState` object.

**Comment by julienbodet** [comment]
> LGTM.

I realized the unit tests are failing when running on macCatalyst. The reason are the following deprecated methods:

image

I'll update the tests to use the newest methods (available for iOS 11+), but because the tests are targeting iOS 8+, I'll need to have some iOS version conditions when archiving/unarchiving in the test classes.

File: Source/iOS/OIDExternalUserAgentCatalyst.h (hunk 7 above)
Comment (julienbodet) [reply Source/iOS/OIDExternalUserAgentCatalyst.h]: Done. I marked this convenience method with `API_UNAVAILABLE(macCatalyst)`.

File: Source/iOS/OIDExternalUserAgentIOS.h [hunk 9]
```diff
@@ -27,6 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 /*! @brief An iOS specific external user-agent that uses the best possible user-agent available
         depending on the version of iOS to present the request.
  */
+API_UNAVAILABLE(macCatalyst)
```
julienbodet: Finally, I replaces the `#if` target statement by this api availability macro. It provides a clearer message to the developer, but still cannot suggest a replacement. macCatalyst being new, this implies AppAuth will be only compatible with Xcode 11+, which is our policy I think.

julienbodet: Finally, I replaced the #if target statement by this api availability macro. It provides a clearer message to the developer, but still cannot suggest a replacement. macCatalyst being new, this implies AppAuth will only be compatible with Xcode 11+, which is our policy I think.

WilliamDenniss: This is great!

Yes, that's our policy. We support old versions of the iOS SDK, but not Xcode (beyond a few months during the update).

WilliamDenniss:
I rebased & squished this in #478, PTAL

WilliamDenniss:
Merged with #478

Shanie1331:
```
In file included from /Users/xxx/ios/Pods/AppAuth/Source/iOS/OIDAuthorizationService+IOS.m:20:
/Users/xxx/ios/Pods/AppAuth/Source/iOS/OIDExternalUserAgentIOS.h:30:1: error: expected ','
API_UNAVAILABLE(macCatalyst)

```

`In file included from /Users/Shashank/Downloads/sept_1/ios/Pods/AppAuth/Source/iOS/OIDAuthorizationService+IOS.m:21:
/Users/Shashank/Downloads/sept_1/ios/Pods/AppAuth/Source/iOS/OIDExternalUserAgentCatalyst.h:28:26: error: expected ','
API_AVAILABLE(macCatalyst(13)) API_UNAVAILABLE(ios)`


Can any one help me with above errors ?

Issue #458, opened 2019-10-23
Different Error Code For Cancelling Authentication Session During Sign Out


LuKenneth:
My issue revolves around signing out. When I launch the sign out session, there is a prompt to cancel it before the external agent window is presented. When this cancel button is tapped, the completion handler for the sign out session presents an error code `OIDErrorCodeUserCanceledAuthorizationFlow -3`. This is expected behavior and perfectly okay with me.

However, if the user continues and then is brought to their identity provider's page in the external agent window, there is another cancel button. When this cancel button is tapped, the completion handler for the sign out session presents the same error code
`OIDErrorCodeUserCanceledAuthorizationFlow -3`. This is an issue for me because the endpoint on the identity provider was already hit, and my identity provider's sign out implementation only requires that endpoint to be visited for sign out to be completed, so tapping the cancel button at this point does not necessarily cancel the session, but it dismisses the window.

For what it's worth, my app is a hybrid web application. I understand that this library is not directed towards hybrid web applications, but this is what I'm stuck with.

**Describe the solution you'd like**
 I would expect a different error code, something like
`OIDErrorCodeExternalUserAgentDismissed`, so that I can respond properly.

The root of the issue is that Apple's error handling for those two cancel events results in the same error code. So somehow I need to distinguish between when the user cancelled the session - did they cancel via the popup prompt asking if they would like to leave their app? Or did they cancel during the external agent window?

I've already looked into this solution, and I have a working solution that will undoubtedly require some advice/direction. I will submit a PR for this. But in summary of the solution - I check what the top view of the app is at the moment the user taps cancel, if it is on the external agent window, then we know they cancelled it at that time.

**Additional context**
Here's screenshots of the two buttons I am referencing in this issue.
![cancel_prompt](https://user-images.githubusercontent.com/11049112/67417564-156f6700-f597-11e9-85c1-4e3958c83fb3.png)
![cancel_window](https://user-images.githubusercontent.com/11049112/67417493-e6f18c00-f596-11e9-9497-a699136815b4.jpeg)

WilliamDenniss:
I agree that distinct error codes would be ideal. However, as you noted, it doesn't appear to be feasible using documented APIs.

I added a few notes on your PR about the limitation of walking the view stack, but also the silver lining that you can use your own user agent implementations with AppAuth without forking.

PR #459, opened 2019-10-23
sends a different error code depending on when the authentication session was cancelled

LuKenneth:
This is a possible solution to issue #458 
https://github.com/openid/AppAuth-iOS/issues/458

I have some doubts about using `UITransitionView`.

1) It is an internal class and cannot be referenced using `isKindOfClass`
2) The view at the time of cancelling may not be guaranteed to be `UITransitionView` (although it seems this way)
3) This requires the view before the pop up prompt to **not** be `UITransitionView` as well...

### DIFF
--- a/Source/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/iOS/OIDExternalUserAgentIOS.m
@@ -100,8 +100,11 @@ NS_ASSUME_NONNULL_BEGIN
         if (callbackURL) {
           [strongSelf->_session resumeExternalUserAgentFlowWithURL:callbackURL];
         } else {
+          int errorCode;
+          NSString *descriptionOfTopView = [[[[[[UIApplication sharedApplication] keyWindow] subviews] lastObject] class] description];
+          errorCode = [descriptionOfTopView compare:@"UITransitionView"] ? OIDErrorCodeExternalUserAgentDismissed : OIDErrorCodeUserCanceledAuthorizationFlow;
           NSError *safariError =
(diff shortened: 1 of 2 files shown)


Source/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -100,8 +100,11 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
         if (callbackURL) {
           [strongSelf->_session resumeExternalUserAgentFlowWithURL:callbackURL];
         } else {
+          int errorCode;
+          NSString *descriptionOfTopView = [[[[[[UIApplication sharedApplication] keyWindow] subviews] lastObject] class] description];
```
WilliamDenniss: Unfortunately, this code is highly susceptible to failing in the future. It is walking a view hierarchy of code we don't own (Apple's implementation of `ASWebAuthenticationSession`).

I don't think we can use this approach.

WilliamDenniss:
Even if we can't merge this code due to it not being future proof, the good news is that you can use AppAuth with a custom OIDExternalUserAgent! Rather than forking AppAuth, you can just add this modified UserAgent to your app (with a new name), and use it in AppAuth.

So you can at least solve this issue for your own app, without needing to maintain your own fork of AppAuth.

WilliamDenniss:
I support the goal here, but I don't think we can use this implementation due to the use of private APIs. Closing as infeasible for now.

PR #469, opened 2019-11-05
[iOS] Allow using third party app to authenticate

r0b0t3d:
Authenticate using app instead of webview.

# How it works
Check if there is `appOAuthUrlScheme` inside `additionalParameters` then build URL for requested app and open it.
# Example
Open `Strava` app for authenticating
```
// builds authentication request
  NSDictionary* parameters = @{ @"appOAuthUrlScheme" : @"strava://oauth/mobile/authorize"};
  OIDAuthorizationRequest *request =
      [[OIDAuthorizationRequest alloc] initWithConfiguration:configuration
                                                    clientId:clientID
                                                clientSecret:clientSecret
                                                      scopes:@[ OIDScopeOpenID, OIDScopeProfile ]
                                                 redirectURL:redirectURI
                                                responseType:OIDResponseTypeCode
                                        additionalParameters:parameters];
```

### DIFF
--- a/Source/OIDAuthorizationRequest.m
+++ b/Source/OIDAuthorizationRequest.m
@@ -353,11 +353,13 @@ NSString *const OIDOAuthorizationRequestCodeChallengeMethodS256 = @"S256";
 }
 
 - (NSURL *)appOAuthUrlScheme {
-  NSString* urlScheme = [self.additionalParameters valueForKey:kAppOAuthUrlScheme];
-  if (urlScheme != NULL) {
-    OIDURLQueryComponent *query = [self queryComponent];
-    NSURL *url = [NSURL URLWithString:urlScheme];
-    return [query URLByReplacingQueryInURL:url];
+  if (self.additionalParameters != NULL) {
+    NSString* urlScheme = [self.additionalParameters valueForKey:kAppOAuthUrlScheme];
+    if (urlScheme != NULL) {
+      OIDURLQueryComponent *query = [self queryComponent];
+      NSURL *url = [NSURL URLWithString:urlScheme];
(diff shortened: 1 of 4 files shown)


WilliamDenniss:
It should be possible to implement an OAuth flow using a native app as the external user agent, without needing to change AppAuth, or introduce a special parameter (`appOAuthUrlScheme` in the PR) for consumption on the client side.

2 approaches to achieve this:

1/
The app tests `canOpenURL`, and if the native app exists, uses the `OIDExternalUserAgentIOSCustomBrowser  CustomBrowserSafari]` in-built AppAuth user agent to open the `strava://...` URI (`CustomBrowserSafari` simply calls `openURL`, so it will open the native app, as long as it's installed and the user confirms).

If `canOpenURl` returns NO (or the user denies the open request), AppAuth is used with the usual `OIDExternalUserAgentIOS` user agent and a `https://` URI as normal.

2/
You write a custom `OIDExternalUserAgent` with same the behavior in this PR.

Side note: I'm not a believer in using a URI parameter to communicate a piece of information internally, URI parameters should be for information exchanged between the client & server in my opinion. If you go down the path of a custom `OIDExternalUserAgent`, I'd suggest using a property of the user agent instead.

Choosing between the two approaches is much of a muchness, option 1 is likely a bit less code. option 2 is more reusable.

I have a TODO to document how to implement a custom `OIDExternalUserAgent` better (#481), as I realize this may not be obvious.

WilliamDenniss:
I added some docs in #483 

Thanks for the contribution. Please file a bug if you find you can't implement this without needing to fork AppAuth, I believe it's possible today through the OIDExternalUserAgent extension point.

Issue #445, opened 2019-09-09
Support token flow


bvirlet:
The Dropbox API supports both the token flow and the code flow.
AppAuth-iOS only supports the code flow.

However, when using the code flow, Dropbox imposes the use of a HTTPS URL for the code flow, and does not support custom schemes. They can only be used for the token flow.
AppAuth-iOS relies on SFAuthenticationSession, which only allows custom URL schemes.

This entirely prevents using AppAuth-iOS to do OAuth with Dropbox.

Supporting the token flow would help a lot in that regard.

PR #474, opened 2019-11-12
Fix build error when build for Mac Catalyst

ANNotunzdY:


WilliamDenniss:
This works. But I think we'll go with #455 instead, which creates a dedicated Catalyst external user agent. The advantage of the approach in #445 is that it allows the two external user agent implementations to evolve separately and cleanly.

PR #480, opened 2019-11-19
target iOS 8.0

dulmandakh:
Xcode showing warning that supported iOS target range is 8.0 to 13.2.99. This PR changes iOS deployment target to be 8.0, or minimum supported version.


WilliamDenniss:
Thanks for the suggestion. It's probably time to drop iOS 7 support, but we need to do it more fully and delete the associated code as well. I've started that in #486

PR #482, opened 2019-11-20
Add Swift code samples to README

WilliamDenniss:


README.md
```diff
@@ -261,6 +357,7 @@ Then, as the port used by the local HTTP server varies, you need to start it
 before building the authorization request, in order to get the exact redirect
 URI to use:
 
+<sub>Objective-C</sub>
```
julienbodet: Any reason to not add a Swift sample here and above?

WilliamDenniss: None in particular. Just ran out of time, and my priority was iOS.

Issue #479, opened 2019-11-19
Unarchive with unarchivedObjectOfClass:fromData:error: fails with Google IdP


julienbodet:
**Describe the bug**
I replaced the deprecated archiving/unarchiving method calls in the example. `unarchivedObjectOfClass:fromData:error:` instead of `unarchivedObjectWithData:` for unarchiving.

While running the app, `OIDAuthorizationRequest.m:147` asserts:
```
2019-11-18 22:40:44.589307-0500 Example-iOS_ObjC[9060:362850] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'The response_type "(null)" isn't supported. AppAuth only supports the "code" or "code id_token" response_type.'
```

By commenting this assert, I get the following error from the unarchive method:
```
Error Domain=NSCocoaErrorDomain Code=4864 "value for key 'NS.objects' was of unexpected class 'NSArray'. Allowed classes are '{(
    OIDServiceDiscovery
)}'." UserInfo={NSDebugDescription=value for key 'NS.objects' was of unexpected class 'NSArray'. Allowed classes are '{(
    OIDServiceDiscovery
)}'.}
```

This seems related to the bug described in #466.

**To Reproduce**
1. Run example app
2. Authorize
3. Kill app
4. Open and see error

**Smartphone (please complete the following information):**
 - Device: iPhone 11 Simulator
 - OS: 13.0

julienbodet:
Please ignore the referencing pull request above, it's an error and unrelated.

julienbodet:
Please ignore the referencing pull request above, it's an error and unrelated.

julienbodet:
Pull request #315 already discussed the same error. The [conclusion](https://github.com/openid/AppAuth-iOS/pull/315#issuecomment-497896461) given by @WilliamDenniss was:

> Steve and I spent a bit of time debugging this. Our conclusion is that the root object of your discovery doc JSON was likely an array, and not a dictionary.

Based on this issue, the previous conclusion made might be wrong and using `unarchivedObjectOfClass:fromData:error:` to unarchive the auth state object could have caused the error observed in #315.

twittemb:
Hi everyone, @julienbodet, @s2ler, @AndrewMcDrew, 

My company is developing a brand new app using AppAuth. And we faced the same issue while using `                NSKeyedUnarchiver.unarchivedObject(ofClass:from:)`.

[...]

PR #487, opened 2019-11-23
Update deprecated archiving methods in the tests

julienbodet:
This PR replaces all archiving/unarchiving deprecated methods in the tests. With these new methods, `OIDRegistrationRequestTests.testSecureCoding` failed because of the decoding error described in #479. So I fixed this too. Lastly, I enabled the tests to run on mac Catalyst.


Source/OIDRegistrationRequest.m
```diff
@@ -141,8 +143,8 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
   NSString *tokenEndpointAuthenticationMethod =
       [aDecoder decodeObjectOfClass:[NSString class]
                              forKey:OIDTokenEndpointAuthenticationMethodParam];
-  NSSet *additionalParameterCodingClasses = [NSSet setWithArray:@[ [NSDictionary class],
-                                                                   [NSString class] ]];
+  additionalParameterCodingClasses = [NSSet setWithArray:@[ [NSDictionary class],
```
WilliamDenniss: Can we introduce a new variable name here rather than re-using the previous one, I feel like it makes it simpler to read that way.

WilliamDenniss:
I'm confused why you're getting that error (though I'm not questioning that you are), and why this code...
```objc
  NSSet *additionalParameterCodingClasses = [NSSet setWithArray:@[ [NSArray<NSURL *> class],
                                                                   [NSURL class] ]];
  NSArray<NSURL *> *redirectURIs = [aDecoder decodeObjectOfClasses:additionalParameterCodingClasses
                                                            forKey:kRedirectURIsKey];
```
https://github.com/openid/AppAuth-iOS/pull/487/files#diff-74f1a1c99ca267845f1642ba1b92f7e0R133-R136

... would fix it.

Isn't: https://github.com/openid/AppAuth-iOS/blob/7e0b5c100d69caac34372efdad34b22b3464ddea/Source/OIDRegistrationRequest.m#L133-L134

The right way to decode:
https://github.com/openid/AppAuth-iOS/blob/7e0b5c100d69caac34372efdad34b22b3464ddea/Source/OIDRegistrationRequest.h#L57

Why does NSURL need to be included as a permitted decoding class? Is what the test is encoding buggy?

Source/OIDRegistrationRequest.m
julienbodet: Yes sure, I'll do that!

julienbodet:
> I'm confused why you're getting that error (though I'm not questioning that you are), and why this code...
> 
> ```objc
>   NSSet *additionalParameterCodingClasses = [NSSet setWithArray:@[ [NSArray<NSURL *> class],
>                                                                    [NSURL class] ]];
>   NSArray<NSURL *> *redirectURIs = [aDecoder decodeObjectOfClasses:additionalParameterCodingClasses
>                                                             forKey:kRedirectURIsKey];
> ```
> 
> https://github.com/openid/AppAuth-iOS/pull/487/files#diff-74f1a1c99ca267845f1642ba1b92f7e0R133-R136
> 
> ... would fix it.

To be honest, I'm confused too. The error message says clearly that it can't decode `NSArray`, so I though adding `NSArray` to the decoding classes could fix it and it does, but I don't really know why!

> Isn't:
> 
> https://github.com/openid/AppAuth-iOS/blob/7e0b5c100d69caac34372efdad34b22b3464ddea/Source/OIDRegistrationRequest.m#L133-L134
> 
> The right way to decode:
> https://github.com/openid/AppAuth-iOS/blob/7e0b5c100d69caac34372efdad34b22b3464ddea/Source/OIDRegistrationRequest.h#L57

That's what I though, unfortunately I have no explanation for that... And Apple documentation doesn't say much about this function and the nested types.

> Why does NSURL need to be included as a permitted decoding class? Is what the test is encoding buggy?

To be sure the test data is not the cause of that bug, I tried reproducing in a Swift Playground and was able to get the same error. I then ask a [question on SO](https://stackoverflow.com/questions/59020621/decoding-array-of-url-objects-using-nskeyedunarchiver) to get an explanation of this behaviour. The only answer I got suggest this is the right way to do. The weird thing is that for arrays of String objects, it works without passing `NSString` as a decoding class.

To me it seems like a bug in the implementation of `NSCoding`/`NSSecureCoding` by `NSArray` or it is intentional but the doc doesn't mention that.

PR #496, opened 2020-01-06
Swift Package Manager support

BalazsSzamody:
- Swift package manifest created
- conditional imports fixed


0xced:
@BalazsSzamody Why did you close this pull request? I think it would be nice to have Swift Package Manager support.

PR #500, opened 2020-01-19
Improve README to reflect actual implementations across iOS versions

0xced:


idelfonsog2:
Thank you for making this README update @0xced Do you know which release does the library supports iOS 12 and 13?

0xced:
According to https://github.com/openid/AppAuth-iOS/releases iOS 12 support was added in version 1.0.0.beta1 and iOS 13 was added in version 1.2.0.

PR #501, opened 2020-01-20
Upgrade to Xcode recommended settings

0xced:
And also set the deployment target at the project level so that all targets inherit the deployment target.


mjenaj:
Get my money

PR #521, opened 2020-04-06
replace # portion of redirect url with ? in order to capture url components

ahmednoursphinx:
…onents

iOS 12.* support for redirect call back not fired in app delegate


WilliamDenniss:
This comes up from time to time. The # fragment is for the token flow which AppAuth doesn't support.

I would suggest implementing this in your own OIDUserAgent instead, if you really need this. You can clone the standard one, modify the URL however you like, before passing it to AppAuth. These user agents are an extension point for AppAuth, so you don't need to fork the library to follow this.

PR #515, opened 2020-03-16
Remove 'UIWebView' in comment to avoid warning when submiting apps to…

julienbodet:
… App Store Connect

Fixes #449.

PR #538, opened 2020-06-03
Update README.md to get rid of ITMS-90809 (UIWebView)

deadjdona:
This change should remove App Store warning - ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ tasks like performing an action with fresh tokens.
 It follows the best practices set out in 
 [RFC 8252 - OAuth 2.0 for Native Apps](https://tools.ietf.org/html/rfc8252)
 including using `SFAuthenticationSession` and `SFSafariViewController` on iOS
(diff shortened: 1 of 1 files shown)


julienbodet:
Hi @deadjdona, thanks for the PR!

Did your app got rejected using AppAuth 1.3.1 or more? We removed a reference to `UIWebView` in a header file in 1.3.1. I'm surprised to see that issue could happen with the README as well.

deadjdona:
Hi Julien.

We use GoogleSignIn library, which use AppAuth 1.2 as dependency, so we still have this issue. Google should update deps...

This PR is my lack of attention, did not noticed that header change, feel free to delete or merge this PR, I don't think readme could cause issues.

Thanks

julienbodet:
This warning has been fixed by #515.

PR #540, opened 2020-06-04
Make ExternalUserAgent default subspec and depend on Core subspec

soule:
- Make the ExternalUserAgent depend on the Core subspec
- Make the ExternalUserAgent the default subspec

(This is in preparation for an EnterpriseUserAgent subspec which would in turn depend on ExternalUserAgent)

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -62,5 +62,5 @@ It follows the OAuth 2.0 for Native Apps best current practice
     externalUserAgent.osx.deployment_target = '10.9'
   end
   
-  s.default_subspec = "ExternalUserAgent"
+  s.default_subspec = 'Core', 'ExternalUserAgent'
 end
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -56,10 +56,8 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
-    externalUserAgent.osx.deployment_target = '10.9'    
-      
+    externalUserAgent.osx.deployment_target = '10.9'      
   end
   
   s.default_subspec = "ExternalUserAgent"
-  
 end
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -31,7 +31,10 @@ It follows the OAuth 2.0 for Native Apps best current practice
   #       classes of AppAuth with tokens on watchOS and tvOS, but currently the
   #       library won't help you obtain authorization grants on those platforms.
 
-  s.platforms    = { :ios => "7.0", :osx => "10.9", :watchos => "2.0", :tvos => "9.0" }
+  s.ios.deployment_target = "7.0"
+  s.osx.deployment_target = "10.9"
+  s.watchos.deployment_target = "2.0"
+  s.tvos.deployment_target = "9.0"
 
   s.source       = { :git => "https://github.com/openid/AppAuth-iOS.git", :tag => s.version }
   s.requires_arc = true
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -47,7 +47,6 @@ It follows the OAuth 2.0 for Native Apps best current practice
     externalUserAgent.dependency 'AppAuth/Core'
     
     externalUserAgent.source_files = "Source/AppAuth.h", "Source/AppAuth/*.{h,m}"
-    externalUserAgent.source_files = "Source/AppAuth/iOS/OIDExternalUserAgentIOSCustomBrowser.*"
     
     # iOS
     externalUserAgent.ios.source_files      = "Source/AppAuth/iOS/**/*.{h,m}"
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -59,7 +59,7 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
-    externalUserAgent.osx.deployment_target = '10.9'      
+    externalUserAgent.osx.deployment_target = '10.9'
   end
   
   s.default_subspec = "ExternalUserAgent"
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -44,8 +44,10 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
   # Subspec for the full AppAuth library, including platform-dependant external user agents.
   s.subspec 'ExternalUserAgent' do |externalUserAgent|
-
-    externalUserAgent.source_files = "Source/*.{h,m}", "Source/AppAuthCore/*.{h,m}", "Source/AppAuth/*.{h,m}"
+    externalUserAgent.dependency 'AppAuth/Core'
+    
+    externalUserAgent.source_files = "Source/AppAuth.h", "Source/AppAuth/*.{h,m}"
+    externalUserAgent.source_files = "Source/AppAuth/iOS/OIDExternalUserAgentIOSCustomBrowser.*"
     
     # iOS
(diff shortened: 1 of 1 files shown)


AppAuth.podspec
```diff
@@ -56,5 +57,9 @@ It follows the OAuth 2.0 for Native Apps best current practice
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
     externalUserAgent.osx.deployment_target = '10.9'    
+
```
yalkhder, line 60: nit: Usually there's no empty line before end.

AppAuth.podspec
```diff
@@ -56,5 +57,9 @@ It follows the OAuth 2.0 for Native Apps best current practice
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
     externalUserAgent.osx.deployment_target = '10.9'    
+      
   end
+  
+  s.default_subspec = "ExternalUserAgent"
+
```
yalkhder, line 64: nit: Usually there's no empty line before end.

AppAuth.podspec
```diff
@@ -55,6 +56,8 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
-    externalUserAgent.osx.deployment_target = '10.9'    
+    externalUserAgent.osx.deployment_target = '10.9'
```
yalkhder, line 59: Remove whitespace.

soule:
@yalkhder @WilliamDenniss I believe that `deployment_targets` is preferred to `platforms` when there are multiple platforms so I changed that: https://guides.cocoapods.org/syntax/podspec.html#platform

I noticed on the CocoaPods guides it states that the subspec will inherit its platforms from the parent if no platforms are specified. However, since the ExternalUserAgent was already defining its own platforms, I understand that this means that ExternalUserAgent only supports ios/macos.

This then leads me to think that the comments should be updated to make it clear that External/EnterpriseUserAgent only support ios/macos, should I add that?

Would like to hear your thoughts, thank you.

AppAuth.podspec
```diff
@@ -44,8 +47,9 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
   # Subspec for the full AppAuth library, including platform-dependant external user agents.
   s.subspec 'ExternalUserAgent' do |externalUserAgent|
-
-    externalUserAgent.source_files = "Source/*.{h,m}", "Source/AppAuthCore/*.{h,m}", "Source/AppAuth/*.{h,m}"
+    externalUserAgent.dependency 'AppAuth/Core'
+    
+    externalUserAgent.source_files = "Source/AppAuth.h", "Source/AppAuth/*.{h,m}"
```
WilliamDenniss, line 52: Not for this PR, but I feel like this directory should probably be renamed `ExternalUserAgent`

AppAuth.podspec
```diff
@@ -31,7 +31,10 @@ It follows the OAuth 2.0 for Native Apps best current practice
   #       classes of AppAuth with tokens on watchOS and tvOS, but currently the
   #       library won't help you obtain authorization grants on those platforms.
 
-  s.platforms    = { :ios => "7.0", :osx => "10.9", :watchos => "2.0", :tvos => "9.0" }
+  s.ios.deployment_target = "7.0"
+  s.osx.deployment_target = "10.9"
+  s.watchos.deployment_target = "2.0"
+  s.tvos.deployment_target = "9.0"
```
WilliamDenniss, line 37: Are you sure this is equivalent? Can you point me to a doc that validates this?

WilliamDenniss: And if it is equivalent, is the line below still needed?
`externalUserAgent.osx.deployment_target = '10.9'`

yalkhder:
> @yalkhder @WilliamDenniss I believe that `deployment_targets` is preferred to `platforms` when there are multiple platforms so I changed that: https://guides.cocoapods.org/syntax/podspec.html#platform
> 
> I noticed on the CocoaPods guides it states that the subspec will inherit its platforms from the parent if no platforms are specified. However, since the ExternalUserAgent was already defining its own platforms, I understand that this means that ExternalUserAgent only supports ios/macos.
> 
> This then leads me to think that the comments should be updated to make it clear that External/EnterpriseUserAgent only support ios/macos, should I add that?
> 
> Would like to hear your thoughts, thank you.

Which comments are you referring to?

AppAuth.podspec
soule: Should this go in a future PR of mine?

AppAuth.podspec
soule: > Are you sure this is equivalent? Can you point me to a doc that validates this?

https://guides.cocoapods.org/syntax/podspec.html#platform : "When supporting multiple platforms you should use deployment_target below instead."

soule: > And if it is equivalent, is the line below still needed?
> `externalUserAgent.osx.deployment_target = '10.9'`

Here https://guides.cocoapods.org/syntax/podspec.html#group_platform it states that "If not defined in a subspec the attributes of this group inherit the value of the parent."

The deployment_targets specified in ExternalUserAgent and EnterpriseUserAgent state that for those subspecs, only osx and ios are supported (From my understanding)

soule:
> Which comments are you referring to?
Lines 29-33

AppAuth.podspec
yalkhder: Yes, have a separate PR for the directory name change.

AppAuth.podspec
WilliamDenniss: Thank you for confirming that, looks good.

I think we can delete the duplicate definitions in the subspecs in that case, there doesn't seem to be any added value by duplicating it.

AppAuth.podspec
```diff
@@ -55,6 +59,8 @@ It follows the OAuth 2.0 for Native Apps best current practice
 
     # macOS
     externalUserAgent.osx.source_files = "Source/AppAuth/macOS/**/*.{h,m}"
-    externalUserAgent.osx.deployment_target = '10.9'    
+    externalUserAgent.osx.deployment_target = '10.9'
```
WilliamDenniss, line 62: Delete this line as it appears not needed, and `externalUserAgent.ios.deployment_target`

AppAuth.podspec
WilliamDenniss: Actually on further reflection, are we sure the EnterpriseUserAgent subspec is not inheriting the tvOS and watchOS deployment targets? The "If not defined in a subspec the attributes of this group inherit the value of the parent." comment refers to the docs for `platform`, not the `deployment_target` so I'm not sure just from reading the docs how it behaves, can you verify with experimentation?

soule: "platform" and "deployment_target" are both subheadings under "Platform" in the docs, so I thought they are equivalents.

I made some test projects. Indeed, if I try to import AppAuth/ExternalUserAgent explicitly in a watchOS or tvOS app podfile, I get an error of this type: "[!] The platform of the target `watchsample` (watchOS 6.2) is not compatible with `AppAuth/ExternalUserAgent (1.4.0)`, which does not support `watchOS`."

However, the behavior is strange when I try to simply import 'AppAuth' in the podfile in watchOS/tvOS.

If `default_subspec = ExternalUserAgent`, then nothing is imported - no Core and no ExternalUserAgent libraries. the imported Pod is empty besides the podspec, readme, license, and xcconfigs.

if `default_subspec = Core, ExternalUserAgent` then when I import 'AppAuth', then the watch/tvOS app imports the Core libraries and AppAuthCore.h, but no ExternalUserAgent files.

Is this okay behavior? I updated the PR to have the two imports.

WilliamDenniss: Thanks for researching this!

The default subspec behavior sounds desirable to me.

WilliamDenniss:
Can you squish this into 1 commit? Then I'll merge it.

Issue #506, opened 2020-01-27
Apple refused our application submission


JesterX666:
After migrating to version 1.3, our application just got refused by Apple Review team.

Reason :

During review, your app installed or launched executable code, which is not permitted on the App Store. Specifically, your app uses the itms-services URL scheme to install an app.

Version 1.3 uses the following code :

+ (instancetype)CustomBrowserChrome {
  // Chrome iOS documentation: https://developer.chrome.com/multidevice/ios/links
  OIDCustomBrowserURLTransformation transform = [[self class] URLTransformationSchemeSubstitutionHTTPS:@"googlechromes" HTTP:@"googlechrome"];
  NSURL *appStoreURL =
  [NSURL URLWithString:@"**itms-apps://itunes.apple.com/us/app/chrome/id535886823**"];
  return [[[self class] alloc] initWithURLTransformation:transform
                                        canOpenURLScheme:@"googlechromes"
                                             appStoreURL:appStoreURL];
}

This is urgent.  If you can fix it fast.

WilliamDenniss:
The way to fix this is to delete `OIDExternalUserAgentIOSCustomBrowser.h` and `OIDExternalUserAgentIOSCustomBrowser.m`, assuming you're not using these (they are not used by default, so likely you are not).

You can clone AppAuth and reference it in your Podfile as a [relative path pod](https://github.com/openid/AppAuth-iOS/blob/7e0b5c100d69caac34372efdad34b22b3464ddea/Examples/Example-iOS_ObjC/Podfile#L6) (like how the samples do it), allowing you to easily delete this file.

Are there any other reports of this type of rejection? If anyone else encounters this, please add your story so we can confirm this is a general trend, and not a one-off.

WilliamDenniss:
Discussed with @petea, a few things:

1/
If `itms-apps://` is no longer allowed (is it?) we should remove it and use a different fallback.

2/
This code isn't used by 99% of users in any case, so it probably should be moved to its own subspec, e.g. "EnterpriseUserAgents" (or removed from the podspec completely and placed in its own "UserAgentExamples" folder) 

If we go the subspec route, we can set the default like so:

default_subspec = 'ExternalUserAgent' so that this new one wouldn't be included by default.

chrismash2:
The way I've directed users to my list of apps on the app store is with https://itunes.apple.com/gb/artist/<name>/<id>?mt=8 (can also be configured to direct to a specific app).

[...]

PR #553, opened 2020-07-10
Replace itms-apps with https in OIDExternalAgentIOSCustomBrowser

soule:
Addresses https://github.com/openid/AppAuth-iOS/issues/506 by removing the use of "itms-apps".

OIDExternalAgentIOSCustomBrowser opens the App Store to install custom browsers if not installed, using the URI itms-apps://. This was replaced by https://, which was tested to instantly launch the App Store in iOS 13. Upon research this behavior may go back as far as iOS 6. In the worst case, https:// would cause a redirect to the browser followed by a redirect to the App Store.


WilliamDenniss:
LGTM

shashi-kiwi:
> LGTM

Hi @WilliamDenniss ,
   I am trying to get latest pull from pod install but still getting ITMS: in code. However, you have merged PR into Master. Can u please help me to avoid apple rejection due ITMS: code.
Please suggest something its very Urgent. Will wait for your reply.
Thanks,
Shashi

PR #564, opened 2020-07-23
[dev-tvos] Add OIDTVTokenRequest class

soule:
Adds OIDTVTokenRequest class, which is a subclass of OIDTokenRequest, to better adhere to the [RFC for TV authentication](https://tools.ietf.org/html/rfc8628#section-3.4). This adds the required `device_code` property which is sent in the request body, which differs from the `code` property for the authorization code in the parent class. Additionally, the grant type for TV authorization is hard-coded in, and several unused fields are dropped from the initializer and body request code.


Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OIDTVTokenRequest : OIDTokenRequest
+
+/*! @brief The device code received from the authorization server.
+    @remarks device_code
+    @see https://tools.ietf.org/html/rfc8628#section-3.4
+ */
+@property(nonatomic, readonly, nullable) NSString *deviceCode;
```
yalkhder, line 34: Is there a situation that this can be nil? If not we should remove `nullable`.

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+
+- (instancetype)init NS_UNAVAILABLE;
```
yalkhder, line 36: Add comment for unavailable methods. E.g. https://github.com/openid/AppAuth-iOS/blob/01131d68346c8ae552961c768d583c715fbe1410/Source/AppAuthCore/OIDTokenResponse.h#L95

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+
+- (instancetype)initWithConfiguration:(OIDServiceConfiguration *)configuration
```
yalkhder, line 38: Add comment.

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+                             clientID:(NSString *)clientID
+                         clientSecret:(nullable NSString *)clientSecret
+                               scopes:(nullable NSArray<NSString *> *)scopes
+                         refreshToken:(nullable NSString *)refreshToken
+                         codeVerifier:(nullable NSString *)codeVerifier
+                 additionalParameters:
+                     (nullable NSDictionary<NSString *, NSString *> *)additionalParameters
+    NS_UNAVAILABLE;
+
+- (instancetype)initWithConfiguration:(OIDServiceConfiguration *)configuration
```
yalkhder, line 51: Add comment.

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+                             clientID:(NSString *)clientID
+                         clientSecret:(nullable NSString *)clientSecret
+                                scope:(nullable NSString *)scope
+                         refreshToken:(nullable NSString *)refreshToken
+                         codeVerifier:(nullable NSString *)codeVerifier
+                 additionalParameters:
+                     (nullable NSDictionary<NSString *, NSString *> *)additionalParameters
+    NS_UNAVAILABLE;
+
+- (instancetype)initWithConfiguration:(OIDTVServiceConfiguration *)configuration
```
yalkhder, line 64: Add docstring comment.

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+                            grantType:(NSString *)grantType
```
yalkhder, line 65: Do we need grantType? Isn't it always going to be urn:ietf:params:oauth:grant-type:device_code?

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+                           deviceCode:(NSString *)deviceCode
+                             clientID:(NSString *)clientID
+                         clientSecret:(NSString *)clientSecret
```
yalkhder, line 68: Is clientSecret needed for Device Authorization access token request?

Source/AppAuthTV/OIDTVTokenRequest.h
```diff
@@ -0,0 +1,79 @@
...
+                 additionalParameters:(NSDictionary<NSString *, NSString *> *)additionalParameters
```
yalkhder, line 69: Should those be nullable, similar to parent class?

Source/AppAuthTV/OIDTVTokenRequest.m
```diff
@@ -0,0 +1,153 @@
...
+    [query addParameter:kGrantTypeKey value:self.grantType];
+  }
+
+  if (self.deviceCode) {
+    [query addParameter:kDeviceCodeKey value:self.deviceCode];
+  }
+
+  [query addParameters:self.additionalParameters];
+
+  return query;
```
yalkhder, line 150: Shouldn't you add the rest of the parameters (at least clientID)?

Source/AppAuthTV/OIDTVTokenRequest.h
soule: It is required, will remove nullable. (https://tools.ietf.org/html/rfc8628#section-3.4)

Source/AppAuthTV/OIDTVTokenRequest.h
soule: Makes sense

Source/AppAuthTV/OIDTVTokenRequest.h
soule: It is not. In the parent class, if clientsecret is present it will do HTTP Basic Authentication otherwise it will just send the client ID in the request body. https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDTokenRequest.m#L277

Source/AppAuthTV/OIDTVTokenRequest.m
soule: Client ID is added in the URLRequest code in the parent; it will add clientID then if not doing http basic authentication ( https://github.com/openid/AppAuth-iOS/blob/master/Source/AppAuthCore/OIDTokenRequest.m#L277 )

Source/AppAuthTV/OIDTVTokenRequest.h
soule: Yes, they should still be nullable, is there anything I need to add here?

soule: Ah, I need to add nullable to the types, will do

Source/AppAuthTV/OIDTVTokenRequest.h
soule: Interestingly, I noticed that in the parent class authorizationCode is nullable, but it is required per the rfc so I'll remove nullable

Source/AppAuthTV/OIDTVTokenRequest.m
```diff
@@ -0,0 +1,154 @@
...
+  }
+  return self;
+}
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+  [super encodeWithCoder:aCoder];
+  [aCoder encodeObject:_deviceCode forKey:kDeviceCodeKey];
+}
+
+- (OIDURLQueryComponent *)tokenRequestBody {
```
WilliamDenniss, line 138: Any reason not to call to super?

Source/AppAuthTV/OIDTVTokenRequest.m
```diff
@@ -0,0 +1,154 @@
...
+  OIDURLQueryComponent *query = [[OIDURLQueryComponent alloc] init];
+
+  if (self.grantType) {
+    [query addParameter:kGrantTypeKey value:self.grantType];
+  }
+
+  if (self.deviceCode) {
```
WilliamDenniss, line 145: I realize this is what we did elsewhere, but I wonder if the null check is needed since this property is nonnull

Source/AppAuthTV/OIDTVTokenRequest.m
```diff
@@ -0,0 +1,154 @@
...
+  [super encodeWithCoder:aCoder];
+  [aCoder encodeObject:_deviceCode forKey:kDeviceCodeKey];
+}
+
+- (OIDURLQueryComponent *)tokenRequestBody {
+  OIDURLQueryComponent *query = [[OIDURLQueryComponent alloc] init];
+
+  if (self.grantType) {
+    [query addParameter:kGrantTypeKey value:self.grantType];
+  }
```
WilliamDenniss, line 143: This could be removed with the call to super, along with the `kGrantTypeKey`

Source/AppAuthTV/OIDTVTokenRequest.m
soule: I could not do it for tokenRequestBody since the method in the parent class is non-public

Source/AppAuthTV/OIDTVTokenRequest.m
soule: Good catch, will fix

PR #571, opened 2020-08-07
Example-tvOS

soule:
Todo: Add README file
Future todo: Discovery?

### DIFF
--- a/Examples/Example-tvOS/Example-tvOS/Info.plist
+++ b/Examples/Example-tvOS/Example-tvOS/Info.plist
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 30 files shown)


Examples/Example-tvOS/Example-tvOS/Info.plist
```diff
@@ -0,0 +1,34 @@
...
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSRequiresIPhoneOS</key>
+	<true/>
```
yalkhder, line 22: This key is for indicating if the app should run on iOS (https://developer.apple.com/documentation/bundleresources/information_property_list/lsrequiresiphoneos?language=objc).

Do we need it?

soule: Thanks, no we don't (interestingly still ran in simulator despite this..)

PR #572, opened 2020-08-12
[dev-tvos] OIDC discovery

soule:


Source/AppAuthCore/OIDServiceDiscovery.h
```diff
@@ -44,6 +44,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property(nonatomic, readonly) NSURL *authorizationEndpoint;
 
+/*! @brief REQUIRED. URL of the OP's OAuth 2.0 Authorization Endpoint.
+    @remarks device_authorization_endpoint
+    @seealso http://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint
+ */
```
WilliamDenniss, line 50: Needs updating. URL is https://tools.ietf.org/html/rfc8628, also this isn't a REQUIRED field.

Source/AppAuthCore/OIDServiceDiscovery.h
```diff
@@ -44,6 +44,12 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property(nonatomic, readonly) NSURL *deviceAuthorizationEndpoint;
```
WilliamDenniss, line 51: `nullable`

Examples/Example-tvOS/Example-tvOS/AppAuthTVExampleViewController.m
```diff
@@ -29,18 +33,6 @@
...
- */
-static NSString *const kTokenEndpoint = @"https://oauth.example.com/token";
-
-/*! @brief User info endpoint.
- */
-static NSString *const kUserInfoEndpoint = @"https://oauth.example.com/userinfo";
-
 /*! @brief NSCoding key for the authorization property.
  */
 static NSString *const kExampleAuthorizerKey = @"authorization";
```
WilliamDenniss, line 38: I wonder if the sample should offer both configuration options like some of the others. The problem with relying on discovery is that not all providers offer it. It wouldn't necessarily need to be a separate button, be a choice the developer makes in the code.

Source/AppAuthCore/OIDServiceDiscovery.h
soule: Just updated it

Source/AppAuthTV/OIDTVAuthorizationService.m
```diff
@@ -23,6 +23,8 @@
 #import "OIDDefines.h"
 #import "OIDErrorUtilities.h"
 #import "OIDURLQueryComponent.h"
+#import "OIDURLSessionProvider.h"
+#import "OIDServiceDiscovery.h"
```
yalkhder, line 27: Fix order (should be alphabetical).

Source/AppAuthTV/OIDTVAuthorizationService.m
```diff
@@ -39,8 +41,94 @@
...
++ (void)discoverServiceConfigurationForIssuer:(NSURL *)issuerURL
+                                   completion:(OIDTVDiscoveryCallback)completion {
+  NSURL *fullDiscoveryURL =
+      [issuerURL URLByAppendingPathComponent:kOpenIDConfigurationWellKnownPath];
+
+  [[self class] discoverServiceConfigurationForDiscoveryURL:fullDiscoveryURL
+                                                 completion:completion];
+}
+
++ (void)discoverServiceConfigurationForDiscoveryURL:(NSURL *)discoveryURL
```
yalkhder, line 62: Seems like there's a lot of shared code with OIDAuthorizationService's method. I think there's an opportunity to rewrite the classes/methods to reduce duplication.

(This is not an actionable item, just general musing).

Source/AppAuthTV/OIDTVServiceConfiguration.h
```diff
@@ -41,11 +41,12 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithAuthorizationEndpoint:(NSURL *)authorizationEndpoint
                                 tokenEndpoint:(NSURL *)tokenEndpoint NS_UNAVAILABLE;
 
-/*! @internal
-    @brief Unavailable. Please use
-        @c initWithAuthorizationEndpoint:TVAuthorizationEndpoint:tokenEndpoint:
```
petea, line 46: Could you also update the comment references to this old selector above?

Examples/Example-tvOS/Example-tvOS/AppAuthTVExampleViewController.m
petea: +1

Source/AppAuthTV/OIDTVAuthorizationService.m
petea, line 27: Check ordering

Examples/Example-tvOS/Example-tvOS/AppAuthTVExampleViewController.m
```diff
@@ -21,9 +21,11 @@
 #import <AppAuth/AppAuthCore.h>
 #import <AppAuth/AppAuthTV.h>
 
-/*! @brief The OIDC issuer from which the configuration will be discovered.
+/*! @brief Indicates whether (TRUE) to discover endpoints from @c kIssuer or (FALSE) to use the
+        @c kTVAuthorizationEndpoint, @c kTokenEndpoint, and @c kUserInfoEndpoint values defined
+        below.
  */
-static NSString *const kIssuer = @"https://issuer.example.com";
+static bool const shouldDiscoverEndpoints = TRUE;
```
yalkhder, line 28: Use `BOOL`. (https://nshipster.com/bool/)

Examples/Example-tvOS/Example-tvOS/AppAuthTVExampleViewController.m
```diff
@@ -21,9 +21,11 @@
 #import <AppAuth/AppAuthCore.h>
 #import <AppAuth/AppAuthTV.h>
 
-/*! @brief The OIDC issuer from which the configuration will be discovered.
+/*! @brief Indicates whether (TRUE) to discover endpoints from @c kIssuer or (FALSE) to use the
```
yalkhder, line 24: nit: No need for the brackets.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -152,47 +154,58 @@ + (NSURL *)googleDiscoveryAuthorizationEndpoint {
...
-      "ndpoint\":\"https://accounts.google.com/o/oauth2/revoke\",\"jwks_uri\":\"https://www.googlea"
-      "pis.com/oauth2/v3/certs\",\"response_types_supported\":[\"code\",\"token\",\"id_token\",\"co"
-      "de token\",\"code id_token\",\"token id_token\",\"code token id_token\",\"none\"],\"subject_"
-      "types_supported\":[\"public\"],\"id_token_signing_alg_values_supported\":[\"RS256\"],\"scope"
-      "s_supported\":[\"openid\",\"email\",\"profile\"],\"token_endpoint_auth_methods_supported\":["
-      "\"client_secret_post\",\"client_secret_basic\"],\"claims_supported\":[\"aud\",\"email\",\"em"
-      "ail_verified\",\"exp\",\"family_name\",\"given_name\",\"iat\",\"iss\",\"locale\",\"name\",\""
-      "picture\",\"sub\"]}";
+    @"{\"issuer\":\"https://accounts.google.com\",\"authorization_endpoint\":\"https://"
+    @"accounts.google.com/o/oauth2/v2/auth\",\"device_authorization_endpoint\":\"https://"
```
WilliamDenniss, line 158: Are there any tests that *don't* include the device_authorization_endpoint? Might be good, since it's optional.

Source/AppAuthTV/OIDTVServiceConfiguration.m
```diff
@@ -44,6 +45,15 @@ - (instancetype)initWithAuthorizationEndpoint:(NSURL *)authorizationEndpoint
                                 tokenEndpoint:(NSURL *)tokenEndpoint
     OID_UNAVAILABLE_USE_INITIALIZER(@selector(initWithTVAuthorizationEndpoint:tokenEndpoint:))
 
+- (instancetype)initWithDiscoveryDocument:(OIDServiceDiscovery *)discoveryDocument {
+  self = [super initWithDiscoveryDocument:discoveryDocument];
+
+  if (self) {
+    _TVAuthorizationEndpoint = [discoveryDocument.deviceAuthorizationEndpoint copy];
```
WilliamDenniss, line 52: How should we handle nil here? What happens if the user attempts to use the OIDTVServiceConfiguration with a nil _TVAuthorizationEndpoint?

UnitTests/OIDServiceDiscoveryTests.m
soule: I think minimumServiceDiscovery does this https://github.com/soule/AppAuth-iOS/blob/dev-tvos-discovery/UnitTests/OIDServiceDiscoveryTests.m#L91

soule: There are some tests that use that i.e testErrorWhenMissingFields

Source/AppAuthTV/OIDTVServiceConfiguration.m
soule: I tested, and it gives an unsupported URL error. I tried to catch this with this check in the service https://github.com/openid/AppAuth-iOS/pull/572/files#diff-0ba4897b100bdc309d5ed2a146f3a331R68

UnitTests/OIDServiceDiscoveryTests.m
soule: I think minimumServiceDiscovery does this https://github.com/soule/AppAuth-iOS/blob/dev-tvos-discovery/UnitTests/OIDServiceDiscoveryTests.m#L91
There are some tests that use that, i.e testErrorWhenMissingFields
(Reposting my reply since it wasn't showing up on conversation page)

Source/AppAuthTV/OIDTVAuthorizationService.m
soule: I reduced the duplication by having [discoverServiceConfigurationForDiscoveryURL:] call the function in OIDAuthorizationService

Source/AppAuthTV/OIDTVAuthorizationService.m
```diff
@@ -60,73 +60,29 @@ + (void)discoverServiceConfigurationForIssuer:(NSURL *)issuerURL
...
-      dispatch_async(dispatch_get_main_queue(), ^{
-        completion(nil, error);
-      });
-      return;
-    }
+                                         completion:(OIDTVDiscoveryCallback)completion {
+  // Call the corresponding discovery method in OIDAuthorizationService
+  [OIDAuthorizationService discoverServiceConfigurationForDiscoveryURL:discoveryURL
+      completion:^(OIDServiceConfiguration * _Nullable configuration, NSError * _Nullable error) {
+    if (configuration) {
```
yalkhder, line 67: Instead of nesting if statements, try to have early exits/returns.

e.g.
```
if (configuration == nil) {
  completion(nil, error);
  return;
}

if (configuration.discoveryDocument.deviceAuthorizationEndpoint == nil {
  // rest of the code
```

Source/AppAuthTV/OIDTVAuthorizationService.m
```diff
@@ -60,73 +60,29 @@ + (void)discoverServiceConfigurationForIssuer:(NSURL *)issuerURL
...
+      if (configuration.discoveryDocument.deviceAuthorizationEndpoint == nil) {
+        NSError *missingEndpointError = [OIDErrorUtilities
+              errorWithCode:OIDErrorCodeInvalidDiscoveryDocument
+            underlyingError:nil
+                description:@"Discovery document does not contain device authorization endpoint."];
+
+        completion(nil, missingEndpointError);
```
yalkhder, line 74: Shouldn't you have a return here so we don't try to create `TVConfiguration` if `deviceAuthorizationEndpoint` is nil?

UnitTests/OIDServiceDiscoveryTests.m
WilliamDenniss: OK, great!

Source/AppAuthTV/OIDTVServiceConfiguration.m
WilliamDenniss: OK, I can get a `OIDServiceDiscovery` object not using your method.

Can you add a `NSLog` here to alert developers to the condition?

Source/AppAuthTV/OIDTVAuthorizationService.h
```diff
@@ -52,6 +60,30 @@ typedef void (^OIDTVAuthorizationCancelBlock)(void);
 /*! @brief Performs authorization flows designed for TVs and other limited input devices.
  */
 @interface OIDTVAuthorizationService : NSObject
+/*! @internal
+    @brief Unavailable. This class should not be initialized.
+ */
+- (instancetype)init NS_UNAVAILABLE;
+
+/*! @brief Convenience method for creating an authorization service configuration from an OpenID
```
WilliamDenniss, line 68: "TV authorization service configuration" ? here and below

WilliamDenniss, line 68: I think it's worth documenting why a developer would use this method, instead of the non-TV AppAuth discovery method and then calling `- (instancetype)initWithDiscoveryDocument:(OIDServiceDiscovery *)discoveryDocument {`.

That reason being (I think), that this method will perform more error checking and is designed for clients doing the TV authorization flow.

Here and below.

Examples/Example-tvOS/Example-tvOS/AppAuthTVExampleViewController.m
```diff
@@ -167,7 +167,7 @@ - (IBAction)signin:(id)sender {
   }
 }
 
--(void) performAuthorizationWithConfiguration:(OIDTVServiceConfiguration *) configuration {
+- (void)performAuthorizationWithConfiguration:(OIDTVServiceConfiguration *) configuration {
```
yalkhder, line 170: Remove space before `configuration`.
```suggestion
- (void)performAuthorizationWithConfiguration:(OIDTVServiceConfiguration *)configuration {
```

PR #611, opened 2020-12-19
Update AppAuth to release 1.4.0

kydonnelly:
We use AppAuth to connect to backup services like Google Drive or Box. The old version uses SFSafariViewController which does not behave well for catalyst.

AppAuth 1.4.0 has specific support for catalyst through ASWebAuthenticationSession as well as some update for iOS 13. I merged our old master branch in, keeping Marc's previous tweaks which you can see in this PR.


Source/AppAuth/iOS/OIDExternalUserAgentIOS.h
```diff
@@ -34,15 +34,15 @@ NS_ASSUME_NONNULL_BEGIN
 API_UNAVAILABLE(macCatalyst)
 @interface OIDExternalUserAgentIOS : NSObject<OIDExternalUserAgent>
 
-- (nullable instancetype)init API_AVAILABLE(ios(11))
+- (nonnull instancetype)init API_AVAILABLE(ios(11))
```
kydonnelly, line 37: this was giving a warning because NSObject init() is non null. Using nullable has no warnings and makes calling code simpler

PR #630, opened 2021-03-31
Token revocation rfc7009

carayolthomas:
This PR implements the [rfc7009](https://tools.ietf.org/html/rfc7009).

Token revocation is available on iOS thanks to the `OIDEndSessionRequest`. However, on tvOS, `AppAuth` doesn't have anything to really logout a user. In such cases, it appears that `AppAuth` should give us (developers) a way to revoke `accessToken` and `refreshToken`.

This first commit is about parsing the `revocation_endpoint` if it exists during the discovery phase.

The second commit implements the `OIDRevokeTokenRequest` and `OIDRevokeTokenResponse` to be able to revoke tokens from the `OIDTVAuthorizationService`.


retendo:
Any progress on this? I would be interested to have that feature in the library.
Are there any recommendations to support the revocation endpoint with the current version?

ChrisProlls:
Up !
No news on that ?

StatusQuo:
We would be happy to have that feature onboard. Any reason it is not merged yet?

PR #633, opened 2021-04-29
macOS: Improve HTML page when authorization completes in the browser

roop:
When using AppAuth in macOS, AppAuth creates a local webserver in the app so that the authorization flow can use the default browser. Once the authorization flow completes, the page shown on the tab used for authorization is currently a very bare HTML page with no styling at all. This PR adds some basic styling that makes the page look better. The same HTML template is used in case authorization succeeded or failed.


roop:
The "swift test" failure in travis-ci seems to be unrelated to this change -- it still fails without these changes.

stefanfuerst:
I like the idea of a nicer HTML page.
However, I would prefer to provide the strings by accessors instead of statics.
This way you cloud subclass OIDRedirectHTTPHandler and show your own **localised** strings to the enduser.

PR #646, opened 2021-06-24
Remove OIDTVTokenRequestTests from sources for AppAuth-iOS scheme.

danblakemore:
AppleTV test file was included in a source target which did not have XCTest dependencies and was causing builds to fail.


zhaiduo:
I delete this too. But it does not work.

`Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_OIDTVTokenRequest", referenced from:
      objc-class-ref in libAppAuth-iOS.a(OIDTVTokenRequestTests.o)
     (maybe you meant: _OBJC_CLASS_$_OIDTVTokenRequestTests)
  "_OBJC_CLASS_$_OIDTVServiceConfiguration", referenced from:
      objc-class-ref in libAppAuth-iOS.a(OIDTVTokenRequestTests.o)
ld: symbol(s) not found for architecture arm64`

danblakemore:
That's what this other PR is for:
https://github.com/openid/AppAuth-iOS/pull/646

I tried to keep them separate because they're conceptually unrelated.

PR #604, opened 2020-11-30
Change iOS deployment target for Xcode 12

davbeck:
This removes the warning about iOS 8 no longer being supported in Xcode 12 while maintaining support for older versions of Xcode. Technically checking the compiler version isn’t an exact 1:1 for checking for Xcode version, but in practice should work in all scenarios.

PR #649, opened 2021-07-15
Support for Xcode 13 beta 3

cabeca:
This PR depends on and includes #604

### DIFF
--- a/Source/AppAuth/iOS/OIDExternalUserAgentIOS.m
+++ b/Source/AppAuth/iOS/OIDExternalUserAgentIOS.m
@@ -163,10 +163,17 @@ NS_ASSUME_NONNULL_BEGIN
       openedUserAgent = YES;
     }
   }
-  // iOS 8 and earlier, use mobile Safari
-  if (!openedUserAgent){
-    openedUserAgent = [[UIApplication sharedApplication] openURL:requestURL];
-  }
+// Xcode 13 beta 3 introduces a breaking change for APIs marked unavailable for iOS extensions
+// https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes
+// The solution is to annotate apis that depend on APIs marked unavailable (sharedApplication, openURL).
+// An easier solution is to remove those APIs in the first place.
+// less than iOS 15, on Xcode 13
+#if __IPHONE_OS_VERSION_MAX_ALLOWED < 150000
+    // iOS 8 and earlier, use mobile Safari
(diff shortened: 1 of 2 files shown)


Source/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -163,10 +163,17 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
-  // iOS 8 and earlier, use mobile Safari
-  if (!openedUserAgent){
-    openedUserAgent = [[UIApplication sharedApplication] openURL:requestURL];
-  }
+// Xcode 13 beta 3 introduces a breaking change for APIs marked unavailable for iOS extensions
+// https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes
+// The solution is to annotate apis that depend on APIs marked unavailable (sharedApplication, openURL).
+// An easier solution is to remove those APIs in the first place.
+// less than iOS 15, on Xcode 13
+#if __IPHONE_OS_VERSION_MAX_ALLOWED < 150000
```
d-date, line 171: IMO, this issue is depends on not iOS 15 but Xcode 13, so this fix is no effect for earlier versions.
By using `NS_EXTENSION_UNAVAILABLE_IOS` macro for whole function, this issue will be fixed.

cabeca: I tried that route first. As soon as you mark this function `NS_EXTENSION_UNAVAILABLE_IOS` , you have to mark all functions depending on this one with `NS_EXTENSION_UNAVAILABLE_IOS` in turn. You will need to do this in 4 places in this library, and the user of this library will have to also mark any of their methods that use one of these newly annotated methods with `NS_EXTENSION_UNAVAILABLE_IOS`

Instead of creating this cascading effect that will eventually end up in the users of this library, I have built upon the fact that if you're using Xcode 12 or above, you will need to use #604 that drops support for iOS 8. If you are using Xcode 13, you're definitely going to have to use #604 , so this code becomes irrelevant:

```swift
    // iOS 8 and earlier, use mobile Safari
     if (!openedUserAgent){
       openedUserAgent = [[UIApplication sharedApplication] openURL:requestURL];
     }
```

This code contains the only two apis that cause the need for `NS_EXTENSION_UNAVAILABLE_IOS`: `sharedApplication` and `openURL` . If we conditionally remove them for Xcode 13, the problem goes away.

PR #661, opened 2021-11-19
Remove platforms declaration to avoid deployment target warnings

CemYil03:
This is required to fix the following warning in Xcode 13.1:
"The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 15.0.99."

PR #673, opened 2021-12-28
Remove IPHONEOS_DEPLOYMENT_TARGET warning by upgrading iOS to 9

minuscorp:


CemYil03:
#661 already covers this.

minuscorp:
Still not merged though

Oliver-Binns:
#661 is a better implementation of this. It would be preferable to merge that instead.

The platform declarations can be removed entirely as they match the minimum version supported by SPM:
https://forums.swift.org/t/xcode12-minimum-deployment-target-and-spm/40467
https://github.com/nanopb/nanopb/pull/585/files

minuscorp:
I saw it, and saw that it wasn't reviewed that's why I posted another one. I really hate warnings in projects.

mattneub:
Please merge this change! We need to get this warning off our plate.

petea:
This was addressed in #690 and released with 1.5.0.  What warning are you seeing?

mattneub:
Yes, I just realized that. The problem is that AppAuth is an indirect dependency so I'm not certain that the version we get is up to us. We're looking into it! Thanks.

PR #679, opened 2022-01-27
prefersEphemeralWebBrowserSession

soledue:
A Boolean value that indicates whether the session should ask the browser for a private authentication session.


petea:
Thanks for the PR!

We're looking at providing this capability via the more general approach of #645.

PR #681, opened 2022-02-01
Use TARGET_OS_OSX for macOS rather than TARGET_OS_MAC

Alex-4-Git:
TARGET_OS_OSX is the one for macOS while TARGET_OS_MAC is for any Apple platform.


petea:
Let's clean up the other references to this macro as well:

https://github.com/openid/AppAuth-iOS/search?q=TARGET_OS_MAC

petea (approved):

PR #488, opened 2019-11-24
Correct nullability annotations in OIDExternalUserAgentIOS

dermaaarkus:
Closes #442 

It has been wrong to declare the initializers as `nullable`, because it conflicted with the implementation of `NSObject`. Please see the description in #442.

From the `NSObject` documentation:

> In terms of nullability, callers can assume that the NSObject implementation of init() does not return nil.

PR #689, opened 2022-03-16
Fix nullability specifier warning UserAgentIOS

andraskadar:
In `OIDExternalUserAgentIOS` we override the default initializer inherited from `NSObject` while also overriding the nullability specifier. In modern Xcode versions this produces a warning.

Ignoring the clang warning removes the warning and states that it was an intended modification.

---

Screenshot from Xcode:
Screen Shot 2022-03-16 at 8 58 22 AM


petea:
This should be addressed by #488

andraskadar:
> This should be addressed by #488

Sounds good, thanks!

PR #702, opened 2022-04-08
Correct nullability annotations in OIDEndSessionRequest

### LINKED ISSUE #565: OIDEndSessionRequest initializers lack correct nullability specifiers
**Are you filing an issue about iOS 12?**
No

**Describe the bug**
`OIDEndSessionRequest.h` brackets its declarations with `NS_ASSUME_NONNULL`, which means that all object-typed parameters will be bridged into Swift as non-optional, unless otherwise specified.

The initializers to create an `OIDEndSessionRequest` allow you to pass in an `idTokenHint:(NSString *)idTokenHint` parameter, which lacks the `nullable` specifier. This means that in Swift, that parameter is bridged as `String`, and not `String?`.

The `@property` for this parameter correctly shows it as `nullable`, and the implementation allows for the underlying ivar to be `nil`, which means the initializer parameter should allow `nil` values as well.

The same applies to the `postLogoutRedirectURL:(NSURL *)postLogoutRedirectURL` and `state:(NSString *)state` parameters.

**Expected behavior**
These initializer parameters are declared as `(nullable NS... *):`

philet:
Correct nullability annotations for `OIDEndSessionRequest` parameters `idTokenHint`, `postLogoutRedirectURL`, and `state`.

Fixes https://github.com/openid/AppAuth-iOS/issues/565

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -42,13 +42,19 @@ It follows the OAuth 2.0 for Native Apps best current practice
   # Subspec for the core AppAuth library classes only, suitable for extensions.
   s.subspec 'Core' do |core|
      core.source_files = "Source/AppAuthCore.h", "Source/AppAuthCore/*.{h,m}"
+     core.resource_bundles = {
+       "AppAuthCore_Privacy" => ["PrivacyInfo.xcprivacy"]
+     }
   end
 
-  # Subspec for the full AppAuth library, including platform-dependant external user agents.
+  # Subspec for the full AppAuth library, including platform-dependent external user agents.
   s.subspec 'ExternalUserAgent' do |externalUserAgent|
     externalUserAgent.dependency 'AppAuth/Core'
     
     externalUserAgent.source_files = "Source/AppAuth.h", "Source/AppAuth/*.{h,m}"
+    externalUserAgent.resource_bundles = {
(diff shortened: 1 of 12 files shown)


petea:
@philet  have you signed the [contributor agreements](https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md)?

philet:
@petea I have signed them last Friday. Probably still needs to be processed on your side.

petea:
@philet confirmed that we have the needed documentation from you, thanks.

petea:
Since this will be a breaking change for the Swift interface, we'll need to make it as part of a major release.

nbransby:
@petea since its only adding the nullability won't existing code still compile meaning its not a breaking change?

pfei5:
We're having to run a fork of this lib because of this issue. Is there anything missing from the PR that we could provide?

xeniabeckel (approved):

matax87:
Any update on this?

trykovyura:
up

MichaelHackett:
@petea: Please provide an update. Is @nbransby correct about it not being a breaking change? I'd rather not have to maintain a fork of the library just to address this one issue, so I'd like to see this finally accepted and merged. It looks to me like it wasn't even changed in the 2.0.0 release, where a breaking change (if it actually is) could have been made.

PR #704, opened 2022-04-22
Fixing the unarchivedObject

maamjadi:
NSCoder is not very good at decoding heterogeneous objects.. With this work the issue with `decodeObject` is fixed as result of which `NSKeyedUnarchiver.unarchivedObject(ofClass: OIDAuthState.self, from: decoded) throws` should work as expected.
This PR is relate to the issue https://github.com/openid/AppAuth-iOS/issues/479

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -20,3 +20,4 @@ Hernan Zalazar <hernan.zalazar@gmail.com> https://github.com/hzalaz
 Joseph Heenan <joseph@emobix.co.uk> https://github.com/jogu
 Julien Bodet <julien.bodet92@gmail.com> https://github.com/julienbodet
 Tobias Schröpf <schroepf@gmail.com> https://github.com/schroepf
+Amin Amjadi <ma.amjadi@outlook.com> https://github.com/maamjadi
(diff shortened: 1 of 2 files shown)


nitrag:
Thank you! I can confirm this is working for me.

```swift
static let userDefaultSuiteName = "group.net.openid.appauth.XYZOAuth"
private var authState: OIDAuthState? {
    didSet {
        if let authState = authState,
            let userDefaults = UserDefaults(suiteName: XYZOAuth.userDefaultSuiteName),
            let data = try? NSKeyedArchiver.archivedData(withRootObject: authState, requiringSecureCoding: true) {
            userDefaults.set(data, forKey: "authState")
            userDefaults.synchronize()
        }
    }
}
```

```swift
let instance = XYZOAuth()
if let data = UserDefaults(suiteName: XYZOAuth.userDefaultSuiteName)?.object(forKey: "authState") as? Data,
   let authStateSaved = try? NSKeyedUnarchiver.unarchivedObject(ofClass: OIDAuthState.self, from: data) {
    instance.authState = authStateSaved
}
return instance
```

petea:
Thanks for the PR.  #479 has been addressed in the 1.6.0 release.

PR #706, opened 2022-05-09
Set prefersEphemeralWebBrowserSession to true 

ruben-ml:
…wser doesn’t share cookies or other browsing data between the authentication session and the user’s normal browser session.


petea:
Thanks for the PR.

We're looking at providing this capability via the more general approach of #645.

PR #710, opened 2022-06-10
Relax redirect URL matching for the forward slash path case

### LINKED ISSUE #446: Trailing slash in URL breaks the auth flow
**Describe the bug**
We are using a third party OpenID provider that appends a slash to the redirect URL, e.g.  `scheme://my-domain.com/?code=...` (note the trailing slash after my-domain.com). When this URL is opened by the browser, the user is taken back to the app but the auth flow gets stuck. Neither a result nor an error is produced. We are using AppAuth iOS Version 1.1.

**Expected behavior**
The AppAuth Android library handles this correctly and is ables to extract the authorization code from the URL, as i think should be the expected behavior.

**Smartphone (please complete the following information):**
 - Device: iPhone 8
 - OS: iOS 12.4
 - Browser: stock browser

**Additional context**
We are using a small wrapper around app auth for flutter (https://github.com/MaikuB/flutter_appauth) which should have no effect on the library's function.

AlexPlekhov:
Based on issue #446 sometimes a trailing slash added to path for redirect url.
Here a fix to make comparison regardless that trailing slash.

This issue affects react-native-app-auth lib:
https://github.com/FormidableLabs/react-native-app-auth/issues/123
https://github.com/FormidableLabs/react-native-app-auth/issues/336

and a few others.

### DIFF
--- a/Source/AppAuthCore/OIDAuthorizationService.m
+++ b/Source/AppAuthCore/OIDAuthorizationService.m
@@ -107,15 +107,17 @@ NS_ASSUME_NONNULL_BEGIN
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
+  NSString *trimmedStdUrlPath = [standardizedURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
+  NSString *stdRedirectUrlPath = [standardizedRedirectURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
+    
 
   return [standardizedURL.scheme caseInsensitiveCompare:standardizedRedirectURL.scheme] == NSOrderedSame
       && OIDIsEqualIncludingNil(standardizedURL.user, standardizedRedirectURL.user)
       && OIDIsEqualIncludingNil(standardizedURL.password, standardizedRedirectURL.password)
       && OIDIsEqualIncludingNil(standardizedURL.host, standardizedRedirectURL.host)
       && OIDIsEqualIncludingNil(standardizedURL.port, standardizedRedirectURL.port)
-      && OIDIsEqualIncludingNil(standardizedURL.path, standardizedRedirectURL.path);
+      && OIDIsEqualIncludingNil(trimmedStdUrlPath, stdRedirectUrlPath);
 }
-
 - (BOOL)shouldHandleURL:(NSURL *)URL {
   return [[self class] URL:URL matchesRedirectionURL:_request.redirectURL];
 }
--- a/Source/AppAuthCore/OIDAuthorizationService.m
+++ b/Source/AppAuthCore/OIDAuthorizationService.m
@@ -107,9 +107,9 @@ NS_ASSUME_NONNULL_BEGIN
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
-  // Some servers adds '/' to the end when there is no 'path'. To relax the equality rules below
-  // were decided to normalize pathes. So, pathes like '' are the same to '/' now.
-  // Read more https://github.com/openid/AppAuth-iOS/issues/446
+  // An empty path may be represented as '' or '/'.  In order to treat these two cases as equivalent,
+  // we normalize a path of '/' to ''.
+  // For context: https://github.com/openid/AppAuth-iOS/issues/446
   NSString *normalizedPath = [standardizedURL.path isEqualToString:@"/"] ? @""
       : standardizedURL.path;
   NSString *normalizedRedirectPath = [standardizedRedirectURL.path isEqualToString:@"/"] ? @""
--- a/Source/AppAuthCore/OIDAuthorizationService.m
+++ b/Source/AppAuthCore/OIDAuthorizationService.m
@@ -107,10 +107,13 @@ NS_ASSUME_NONNULL_BEGIN
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
+  // Some servers adds '/' to the end when there is no 'path'. To relax the equality rules below
+  // were decided to normalize pathes. So, pathes like '' are the same to '/' now.
+  // Read more https://github.com/openid/AppAuth-iOS/issues/446
   NSString *normalizedPath = [standardizedURL.path isEqualToString:@"/"] ? @""
-    : standardizedURL.path;
+      : standardizedURL.path;
   NSString *normalizedRedirectPath = [standardizedRedirectURL.path isEqualToString:@"/"] ? @""
-    : standardizedRedirectURL.path;
+      : standardizedRedirectURL.path;
 
   return [standardizedURL.scheme caseInsensitiveCompare:standardizedRedirectURL.scheme] == NSOrderedSame
       && OIDIsEqualIncludingNil(standardizedURL.user, standardizedRedirectURL.user)
--- a/Source/AppAuthCore/OIDAuthorizationService.m
+++ b/Source/AppAuthCore/OIDAuthorizationService.m
@@ -107,17 +107,19 @@ NS_ASSUME_NONNULL_BEGIN
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
-  NSString *trimmedStdUrlPath = [standardizedURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
-  NSString *stdRedirectUrlPath = [standardizedRedirectURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
-    
+  NSString *normalizedPath = [standardizedURL.path isEqualToString:@"/"] ? @""
+    : standardizedURL.path;
+  NSString *normalizedRedirectPath = [standardizedRedirectURL.path isEqualToString:@"/"] ? @""
+    : standardizedRedirectURL.path;
 
   return [standardizedURL.scheme caseInsensitiveCompare:standardizedRedirectURL.scheme] == NSOrderedSame
       && OIDIsEqualIncludingNil(standardizedURL.user, standardizedRedirectURL.user)
       && OIDIsEqualIncludingNil(standardizedURL.password, standardizedRedirectURL.password)
       && OIDIsEqualIncludingNil(standardizedURL.host, standardizedRedirectURL.host)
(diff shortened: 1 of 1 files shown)


petea:
Thanks for the contribution @AlexPlekhov, could you sign the [contributor agreements](https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md) so we can get this merged?

AlexPlekhov:
Hi @petea, I've already done it immediately after creating PR. I hope you got email notifications. If didn't I'm ready to send those PDFs via email.

AlexPlekhov:
> Thanks for the contribution @AlexPlekhov, could you sign the [contributor agreements](https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md) so we can get this merged?

Hi @petea, could you confirm that everything OK? I followed the steps in the [contributor agreements](https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md). Please, write me back if I missed anything and still have to do something.

ntherning:
Would be great to have this PR merged as it fixes an issue with Azure AD we have been experiencing. The Azure Portal gives us a redirect URI which looks like `msauth.com.example.app://auth` to use with our iOS app. But when the redirect back to our app happens they have appended a slash which breaks AppAuth's redirect URI validation. With the changes in this PR applied AppAuth happily accepts the redirect URI and proceeds.

AlexPlekhov:
Hi all, @petea,

I've resigned all docs as required by Mike Leszcz, was added to a WG, got a subscription email.
But it's still in review. Have I missed anything again?

developerfromjokela (approved):

developerfromjokela:
I think this should be merged, as this is mission critical for functioning apps that rely on third party services and thus cannot influence how the backend behaves.

Source/AppAuthCore/OIDAuthorizationService.m
```diff
@@ -107,15 +107,17 @@ - (void)cancelWithCompletion:(nullable void (^)(void))completion {
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
+  NSString *trimmedStdUrlPath = [standardizedURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
+  NSString *stdRedirectUrlPath = [standardizedRedirectURL.path stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"/"]];
```
petea, line 111: Rather than trimming forward slashes from both ends of the paths and comparing, I think we should look at a narrower approach that modifies the existing check to consider a path value of "/" as equal to an empty path of "".  One way to achieve this would be to normalize "/" to the empty path "".  This would look something like:

```
// For this URL comparison, normalize path values of "/" to "".
NSString *normalizedPath =
    [standardizedURL.path isEqualToString:@"/"] ? @"" : standardizedURL.path;
NSString *normalizedRedirectPath =
    [standardizedRedirectURL.path isEqualToString:@"/"] ? @"" : standardizedRedirectURL.path;

...

return
...
    && OIDIsEqualIncludingNil(normalizedPath, normalizedRedirectPath);
```

petea (changes requested):

petea:
@developerfromjokela note that there are workarounds for this situation discussed in #446.

Source/AppAuthCore/OIDAuthorizationService.m
AlexPlekhov: > Rather than trimming forward slashes from both ends of the paths and comparing, I think we should look at a narrower approach that modifies the existing check to consider a path value of "/" as equal to an empty path of "". One way to achieve this would be to normalize "/" to the empty path "". This would look something like:
> 
> ```
> // For this URL comparison, normalize path values of "/" to "".
> NSString *normalizedPath =
>     [standardizedURL.path isEqualToString:@"/"] ? @"" : standardizedURL.path;
> NSString *normalizedRedirectPath =
>     [standardizedRedirectURL.path isEqualToString:@"/"] ? @"" : standardizedRedirectURL.path;
> 
> ...
> 
> return
> ...
>     && OIDIsEqualIncludingNil(normalizedPath, normalizedRedirectPath);
> ```

Thank you for suggestion @petea . Not sure that this approach will help in cases, when a path is not just a trailing slash.
So, if a path is 'foo' and the server side will add a trailing slash, we'll compare 'foo' and 'foo/' and get the same issue again.
Am I missed anything?

For my particular case you suggestion will perfectly work, because my app doesn't have any path in callback url. But I'm still not sure in general case.

AlexPlekhov:
> @developerfromjokela note that there are workarounds for this situation discussed in #446.

Yep, we already did the same. But wants to help the community to not face with it again.

davidkaya:
Hi, will this get merged? We encountered this issue just today. We have support for multiple IdPs that support OIDC and customer wanted to use ADFS but it is not working cause ADFS appends `/` to the redirect uri 🤷🏽‍♂️

Source/AppAuthCore/OIDAuthorizationService.m
petea: The empty path case ("" or "/") is the only one in which the two URLs are arguably equivalent.  Adding a trailing slash to a non-empty path component ("/path" -> "/path/") changes the resource to which the URL points, so I don't think we should equate the two.  Luckily, in practice, we seem to be dealing only with the empty path case here.

petea:
@davidkaya note that there is a straightforward workaround for this issue that is detailed in #446.

Source/AppAuthCore/OIDAuthorizationService.m
AlexPlekhov: Ok. Your suggestion is already applied in this PR. My senior colleagues absolutely agree with you as well.

Source/AppAuthCore/OIDAuthorizationService.m
```diff
@@ -107,13 +107,17 @@ - (void)cancelWithCompletion:(nullable void (^)(void))completion {
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
+  NSString *normalizedPath = [standardizedURL.path isEqualToString:@"/"] ? @""
+    : standardizedURL.path;
+  NSString *normalizedRedirectPath = [standardizedRedirectURL.path isEqualToString:@"/"] ? @""
+    : standardizedRedirectURL.path;
```
petea, line 113: Same as above.

petea (changes requested):

Source/AppAuthCore/OIDAuthorizationService.m
```diff
@@ -107,13 +107,17 @@ - (void)cancelWithCompletion:(nullable void (^)(void))completion {
 + (BOOL)URL:(NSURL *)URL matchesRedirectionURL:(NSURL *)redirectionURL {
   NSURL *standardizedURL = [URL standardizedURL];
   NSURL *standardizedRedirectURL = [redirectionURL standardizedURL];
+  NSString *normalizedPath = [standardizedURL.path isEqualToString:@"/"] ? @""
+    : standardizedURL.path;
```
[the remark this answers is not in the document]
AlexPlekhov: Ok. Just for me, please, look at lines 628, 632 and 636.
Could you explain why in the fist of them an indent is 4, but two last is 2? Is there any logic?

DelcoigneYves:
Can we get this PR moving again? I think the only remaining question is a missing comment? Would help in our case (using [the Flutter AppAuth library](https://pub.dev/packages/flutter_appauth) where this is an issue as well (and mentioned workaround does not apply in our case)

AlexPlekhov:
@DelcoigneYves 
I did everything I could. Comments, formats - everything has been in place already.

DelcoigneYves:
Then poking @petea for a merge 💯

petea (changes requested):

AlexPlekhov:
Hi @petea, sorry to trouble you. I've got e massages that some tests are failed. But when I look into results it's more about macs machines rather than real failures.

In xCode it all looks OK.
Screenshot 2022-10-27 at 12 51 52

Do you have any idea what I can do to solve this?

AlexPlekhov:
@DelcoigneYves  @petea @davidkaya 
Hi folks!
I give up. I see only resistance against this PR and ignorance.
May be you will be more successful to resolve this issue on your way.
I will close it soon.

DelcoigneYves:
@petea LGTM! Can we please now merge it? 🙂 https://github.com/openid/AppAuth-iOS/blob/master/CONTRIBUTING.md#pull-request-reviews

pawellaskowski:
Hi @petea, is this going to get merged?

icanci01:
I’m encountering a similar issue in a React Native application that uses AppAuth on iOS. Interestingly, this works fine on Android, but iOS fails due to strict redirect URL matching.

The problem is that I have no control over the Identity Provider (IdP). The IdP rejects the client if I register the redirect URL with a trailing /, so I must use the version without the trailing slash. However, the IdP does append a trailing / when returning the authorization response, causing AppAuth’s strict path comparison to reject the redirect.

Requesting the IdP to change its behavior isn’t feasible, as adding a trailing slash for my client would risk breaking existing integrations.

Could we consider making the redirect URI/path matching more relaxed, or at least providing a configuration option where developers can specify alternative acceptable redirect URLs (e.g., allow both myapp://callback and myapp://callback/)?

The suggested workarounds from previous discussions didn’t solve this issue for me. I resorted to creating a custom patch that’s applied automatically during pod install, but this approach is fragile — any upstream file structure change in AppAuth forces me to maintain and update the patch manually.

I even tried intercepting/replacing the matching method at runtime, but had no success despite multiple attempts.

Could we please get an update on this Pull Request, or some guidance on a cleaner long-term solution? A small change upstream would greatly improve AppAuth’s flexibility for these kinds of real-world IdP inconsistencies.

developerfromjokela:
> I’m encountering a similar issue in a React Native application that uses AppAuth on iOS. Interestingly, this works fine on Android, but iOS fails due to strict redirect URL matching.
> 
> 
> 
> The problem is that I have no control over the Identity Provider (IdP). The IdP rejects the client if I register the redirect URL with a trailing /, so I must use the version without the trailing slash. However, the IdP does append a trailing / when returning the authorization response, causing AppAuth’s strict path comparison to reject the redirect.
> 
> 
> 
> Requesting the IdP to change its behavior isn’t feasible, as adding a trailing slash for my client would risk breaking existing integrations.
> 
> 
> 
> Could we consider making the redirect URI/path matching more relaxed, or at least providing a configuration option where developers can specify alternative acceptable redirect URLs (e.g., allow both myapp://callback and myapp://callback/)?
> 
> 
> 
> The suggested workarounds from previous discussions didn’t solve this issue for me. I resorted to creating a custom patch that’s applied automatically during pod install, but this approach is fragile — any upstream file structure change in AppAuth forces me to maintain and update the patch manually.
> 
> 
> 
> I even tried intercepting/replacing the matching method at runtime, but had no success despite multiple attempts.
> 
> 
> 
> Could we please get an update on this Pull Request, or some guidance on a cleaner long-term solution? A small change upstream would greatly improve AppAuth’s flexibility for these kinds of real-world IdP inconsistencies.

Seems rather hopeless trying to get this merged, 2 years oh boy. maybe starting a fork with the patch is worth it?

icanci01:
@developerfromjokela Yeah, starting a fork with the patch is definitely an option, but it comes with a few drawbacks. Maintaining a fork means I have to keep it in sync with upstream, manually merge updates, and ensure compatibility with future iOS/React Native changes. It works as a quick fix sure.. but I don't want to have another burden... I think for now patching the file during pod install might be the easiest fix for me.. :) will just have to make sure with future updates to update my patch file...

I think the issue with this pull request is that the reviewer @petea seems to not be active anymore..

Can we have someone else that is actively maintaining this to have a look?

PR #721, opened 2022-09-09
Correct OIDServiceDiscovery's NSCoding implementation

petea:
Provide our own implementation of `NSCoding` for `OIDServiceDiscovery` rather than delegating to `NSDictionary` and maintain support for decoding the old encoding for backward compatibility.  Ensure that OIDServiceConfiguration can successfully decode the old encoding when `requiresSecureCoding` is enabled.  Test both new and old encodings with `requiresSecureCoding` enabled.

### DIFF
--- a/UnitTests/OIDServiceDiscoveryTests.m
+++ b/UnitTests/OIDServiceDiscoveryTests.m
@@ -30,13 +30,35 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wgnu"
 
-// Define a subclass of @c OIDServiceDiscovery that provides the old implementation of
-// encodeWithCoder:.
-@interface OIDServiceDiscoveryOld : OIDServiceDiscovery
+// Define a subclass of @c OIDServiceDiscovery that provides the old NSCoding decoding
+// implementation.
+@interface OIDServiceDiscoveryOldDecoding : OIDServiceDiscovery
+@end
+
+@implementation OIDServiceDiscoveryOldDecoding
+
++ (BOOL)supportsSecureCoding {
+  return YES;
+}
+
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
+  NSError *error;
+  NSDictionary *dictionary = [[NSDictionary alloc] initWithCoder:aDecoder];
+  self = [self initWithDictionary:dictionary error:&error];
+  if (error) {
+    return nil;
+  }
+  return self;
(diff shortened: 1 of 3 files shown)


UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -411,8 +425,58 @@ - (void)testSecureCoding {
...
+  OIDServiceDiscovery *unarchived;
+  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    unarchived = [NSKeyedUnarchiver unarchivedObjectOfClass:[OIDServiceDiscovery class]
+                                                   fromData:data
+                                                      error:&error];
+  } else {
+    unarchived = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  }
+
+  XCTAssertEqualObjects(discovery.discoveryDictionary, unarchived.discoveryDictionary, @"");
```
danblakemore, line 446: It's probably unnecessary to specify a string here as the automatic error message is self-explanatory.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -30,6 +30,20 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wgnu"
 
+// Define a subclass of @c OIDServiceDiscovery that provides the old implementation of
+// encodeWithCoder:.
+@interface OIDServiceDiscoveryOld : OIDServiceDiscovery
+
+@end
+
+@implementation OIDServiceDiscoveryOld
```
danblakemore, line 39: Consider also testing the case where serialized service discovery objects are cached and shared between app groups.  Specifically: what happens if a non-updated member of an app group tries to read a cache containing the new serialization?

danblakemore (commented):
See comments for changes.

UnitTests/OIDServiceDiscoveryTests.m
petea: Removed the empty string for the new tests.

UnitTests/OIDServiceDiscoveryTests.m
petea: An error was generated in the old decoding the new scenario.  I've now added forward compatibility and a test.

danblakemore (approved):
Nice!  Even better than not crashing.

PR #747, opened 2023-01-18
Increase swift version to 5.7

osrufung:
The reason I need this upgrade is to be able to use `moduleAliases` in my Package when using AppAuth dependent packages like [GoogleSigIn](https://github.com/google/GoogleSignIn-iOS/blob/main/Package.swift) 

here is the explanation of this issue:
https://github.com/apple/swift-package-manager/issues/5762

cc @mdmathias

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


MapaX (approved):

ankushkushwaha (approved):

ankushkushwaha:
Can you please resolve emerge conflicts with base branch

osrufung:
Not needed anymore, as "moduleAliases" was not the solution we expected.😅

Issue #665, opened 2021-11-30
EndSessionRequest 


iosdev-mk:
**Describe the bug**

OIDEndSessionRequest on iOS can not be instantiated without idTokenHint. 
When using an empty String as idTokenHint, the Keykloak Server shows "Session not Active" an does not destroy the session.

**OIDEndSessionRequest** property **idTokenHint** is marked es 'nullable' but not in the init-Method:

**iOS API OIDEndSessionRequest:**

`@property(nonatomic, readonly, **nullable**) NSString *idTokenHint;

- (instancetype)
    initWithConfiguration:(OIDServiceConfiguration *)configuration
              idTokenHint:(NSString *)idTokenHint
    postLogoutRedirectURL:(NSURL *)postLogoutRedirectURL
     additionalParameters:(nullable NSDictionary<NSString *, NSString *> *)additionalParameters;`

On Android, its possible to create an EndSession-Request without idToken and logout successfully on our Keykloak Server. 

**To Reproduce**
Create EndSessionRequest in Swift
Swift API reqires to specifiy an idTokenHint
Using "" as idTokenHint
Keykloak shows "Session not Active"

**Expected behavior**
The OIDEndSessionRequest Initializer can be called without idTokenHint.

**Screenshots**
<img width="869" alt="Code_EndSessionRequest" src="https://user-images.githubusercontent.com/87985765/144034539-4d0fc9f0-1d91-4559-970f-26b44567a634.png">

![keykloak_sessionnotactive](https://user-images.githubusercontent.com/87985765/144035231-f2f0a557-43c4-455e-90b8-39fc7c2eeaae.png)

**Smartphone (please complete the following information):**
 - Device: iOS Device
 - OS: iOS15
 - Browser SafariViewController
 - AppAuth Version 1.4

iosdev-mk:
To clarify the problem:
We are using the scope "profile" and not "openid", so we don't get an idToken after the authentication. 
Currently we can not start an EndSessionRequest without idToken in Swift, because the the property is not nullable in the initializer.

Neeraj204:
Hi, @iosdev-mk. did you get it solved? In my case, the logout is working, but I am not getting any response or error from the `OIDAuthorizationService.present(currentEndSessionRequest!, externalUserAgent: currentAuthSession!) { [weak self] response, error in`


The following is the full code for logout:

`func logout(from viewController: UIViewController, completion: @escaping (Bool) -> Void) {
        guard let authState = self.authState,
[...]

PR #727, opened 2022-10-06
Fix nullability annotation for -[OIDExternalUserAgentIOS init]

tiwoc:
This PR fixes this build warning:
> Source/AppAuth/iOS/OIDExternalUserAgentIOS.h:37:1: warning build: conflicting nullability specifier on return types, 'nullable' conflicts with existing specifier 'nonnull'

Unfortunately, this warning is misleading. `-[NSObject init]` doesn't have any nullability annotation, which is equivalent to `null_unspecified`, but `-[OIDExternalUserAgentIOS init]` has the conflicting `nullable` annotation.

The straightforward solution would be to remove the annotation here to get in sync with the base class, but this would be incorrect: In line 34, `NS_ASSUME_NONNULL_BEGIN` is used, which would make the annotation default to `nonnull` instead of `null_unspecified`, which is why `null_unspecified` has to be used explicitly.

To verify the fix, run the iOS tests in this repository with and without this PR. With this PR applied, the warning will no longer show up.

Fixes #442

PR #753, opened 2023-02-17
Update OIDEndSessionRequest method signature to match nullability on type

mt00chikin:
This updates the idTokenHint nullability annotation in OIDEndSessionRequest's initializers to match the property attributes on the type, so that Swift clients can use these methods more easily. 
`idTokenHint` is declared as nullable on OIDEndSessionRequest and the implementation of this class already accounts for the possibility of the property being nil. However, the initializer methods omit the nullable annotation for the idTokenHint parameter. 
In Objective-C based clients this does not present a problem, since the compiler will allow nil to be passed in the init method. Swift clients, however, will fail to compile because the Swift interface treats idTokenHint as a `String` and not a `String?`. 

This results in the following updated Swift interface:
```
/** @brief Creates an authorization request with opinionated defaults (a secure @c state).
        @param configuration The service's configuration.
        @param idTokenHint The previously issued ID Token
        @param postLogoutRedirectURL The client's post-logout redirect URI.
            callback.
        @param additionalParameters The client's additional authorization parameters.
    */
    public convenience init(configuration: OIDServiceConfiguration, idTokenHint: String?, postLogoutRedirectURL: URL, additionalParameters: [String : String]?)

    
    /** @brief Designated initializer.
        @param configuration The service's configuration.
        @param idTokenHint The previously issued ID Token
        @param postLogoutRedirectURL The client's post-logout redirect URI.
        @param state An opaque value used by the client to maintain state between the request and
            callback.
        @param additionalParameters The client's additional authorization parameters.
     */
    public init(configuration: OIDServiceConfiguration, idTokenHint: String?, postLogoutRedirectURL: URL, state: String, additionalParameters: [String : String]?)
```

This PR resolves the issue reported in https://github.com/openid/AppAuth-iOS/issues/665

### DIFF
--- a/Source/AppAuthCore/OIDEndSessionRequest.h
+++ b/Source/AppAuthCore/OIDEndSessionRequest.h
@@ -75,7 +75,7 @@ NS_ASSUME_NONNULL_BEGIN
 */
 - (instancetype)
     initWithConfiguration:(OIDServiceConfiguration *)configuration
-              idTokenHint:(NSString *)idTokenHint
+              idTokenHint:(nullable NSString *)idTokenHint
     postLogoutRedirectURL:(NSURL *)postLogoutRedirectURL
      additionalParameters:(nullable NSDictionary<NSString *, NSString *> *)additionalParameters;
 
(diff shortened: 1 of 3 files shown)


mt00chikin:
FYI, I have started the process of executing the agreements per the CONTRIBUTING guidelines

radutx (approved):

radutx (approved):

ir-fuel:
Will this ever get merged?

mt00chikin:
Accomplished by https://github.com/openid/AppAuth-iOS/pull/727

Issue #759, opened 2023-03-28
XCode 14.3 Beta cant archieve app because of AppAuth.framework pod


christopherarm:
**Describe the bug**

I recently installed XCode 14.3 Beta to build my Apps for iOS 16.4. I use flutter for development.
Now i get this error when i try to build the ipa or archieve the app via XCode:

`mkdir -p /Users/chris/Library/Developer/Xcode/DerivedData/Runner-hjyvfkwvhrjdsnelaqvdpdxpoyit/Build/Intermediates.noindex/ArchiveIntermediates/Runner/BuildProductsPath/Release-iphoneos/Runner.app/Frameworks
Symlinked...
rsync --delete -av --filter P .*.?????? --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" --filter "- Headers" --filter "- PrivateHeaders" --filter "- Modules" "../../../IntermediateBuildFilesPath/UninstalledProducts/iphoneos/AppAuth.framework" "/Users/chris/Library/Developer/Xcode/DerivedData/Runner-hjyvfkwvhrjdsnelaqvdpdxpoyit/Build/Intermediates.noindex/ArchiveIntermediates/Runner/InstallationBuildProductsLocation/Applications/Runner.app/Frameworks"
building file list ... rsync: link_stat "/Users/chris/Downloads/tuduu_app/ios/../../../IntermediateBuildFilesPath/UninstalledProducts/iphoneos/AppAuth.framework" failed: No such file or directory (2)
done`

**To Reproduce**
1. Create flutter project with XCode 14.3 Beta installed
2. Run 'flutter build ipa' or run Product>Archieve in XCode

**Expected behavior**
The build should run fine and generate the ipa file.

**Screenshots**
![image](https://user-images.githubusercontent.com/90921702/228348666-5ba5a706-88a4-4a69-9398-34ab86a15044.png)

**Environment**
- Device: [ iPhone 12 ]
- OS: [ iOS 16.4 Beta ]
- XCode 13.4 Beta

junytse:
I meet a similar issue when using `carthage bootstrap` with Xcode 14.3. The log reports a deployment target warning and an error about missing `...`:

```
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphoneos.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)

[...]

PR #761, opened 2023-04-06
Bump iOS minimum version in podspec to support Xcode 14.3

mdmathias:
Fixes https://github.com/openid/AppAuth-iOS/issues/759

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -31,7 +31,7 @@ It follows the OAuth 2.0 for Native Apps best current practice
   #       classes of AppAuth with tokens on watchOS and tvOS, but currently the
   #       library won't help you obtain authorization grants on those platforms.
 
-  s.ios.deployment_target = "7.0"
+  s.ios.deployment_target = "9.0"
   s.osx.deployment_target = "10.9"
   s.watchos.deployment_target = "2.0"
   s.tvos.deployment_target = "9.0"
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -32,7 +32,7 @@ It follows the OAuth 2.0 for Native Apps best current practice
   #       library won't help you obtain authorization grants on those platforms.
 
   s.ios.deployment_target = "9.0"
-  s.osx.deployment_target = "10.9"
+  s.osx.deployment_target = "10.12"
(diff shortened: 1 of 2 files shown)


AppAuth.podspec
```diff
@@ -31,7 +31,7 @@ It follows the OAuth 2.0 for Native Apps best current practice
   #       classes of AppAuth with tokens on watchOS and tvOS, but currently the
   #       library won't help you obtain authorization grants on those platforms.
 
-  s.ios.deployment_target = "7.0"
+  s.ios.deployment_target = "9.0"
   s.osx.deployment_target = "10.9"
```
paulb777, line 35: Probably need 10.12 here to be consistent with 2015 releases.

jmagman: The `libarclite_iphoneos` issue reported in the comment on https://github.com/openid/AppAuth-iOS/issues/759#issuecomment-1487880042 reproduces on macOS when the deployment target is < macOS 10.11 so this should be at least 10.11 to actually fix that linker issue, if not higher.

mdmathias: Thanks for the catch. I have updated this. Podspec is linting locally. Publishing this PRs draft.

paulb777 (approved):
LGTM. 

Remember to also bump the podspec version to 1.6.1 or 1.7.0 before publishing to CocoaPods trunk. My recommendation without full context would be 1.6.1 to make it as seamless as possible for clients to get the update.

mdmathias:
Thanks @paulb777 for the reminder. The last time we bumped a minimum OS version, we did a minor release.

https://github.com/openid/AppAuth-iOS/commit/de08e7847ec9bbc064eda107399d20b423c0e076

I'm inclined to stick with that precedent. Thoughts?

paulb777:
If the version bump affected Xcodes that are still currently be used, I would agree that it needs to be a minor version update. Since this is changing from a long ago version to a still unsupported version, I think a patch version update is better so that clients that lock to a minor version also get the fix without a respin of their podspec.

jmagman (approved):
LGTM, thanks!

ch40w31 (approved):

PR #776, opened 2023-06-29
Add custom browser support for Edge

Liam212:
Implement out of the box custom browser support for microsoft edge


BoikoArtem30091994:
Hello. Can someone accept these changes so that it is still available?

PR #770, opened 2023-04-28
introduce addtional http headers to OIDTokenRequest

break2k:
### Problem

My IDM requests an additional header from token requests we are currently unable to send with the AppAuth SDK.

### Solution

Extend `OIDTokenRequest` similar to `additionalParameters` to use `additionalHeaders`.

### Background

In order to fulfill our demands for quality control inside our corporation, we are adding some generated UUID for the session and the client that we can use for debugging.

Unfortunately, I am unable to share those internal documents.

PR #798, opened 2023-10-20
fixed wrong variable for additional http headers in OIDTokenRequest (#770)

GravityByte:
This update fixes wrong variable copy-pasted code for additional http headers in OIDTokenRequest.

### DIFF
--- a/Source/AppAuthCore/OIDTokenRequest.m
+++ b/Source/AppAuthCore/OIDTokenRequest.m
@@ -330,7 +330,7 @@ static NSString *const kAdditionalHeadersKey = @"additionalHeaders";
   }
   
   for (id header in _additionalHeaders) {
-    [URLRequest setValue:httpHeaders[header] forHTTPHeaderField:header];
+    [URLRequest setValue:_additionalHeaders[header] forHTTPHeaderField:header];
   }
 
   return URLRequest;
--- a/UnitTests/OIDTokenRequestTests.m
+++ b/UnitTests/OIDTokenRequestTests.m
@@ -56,6 +56,14 @@ static NSString *const kTestAdditionalHeaderKey = @"B";
  */
 static NSString *const kTestAdditionalHeaderValue = @"2";
 
+/*! @brief Test key for the @c additionalHeaders property.
+ */
+static NSString *const kTestAdditionalHeaderKey2 = @"C";
+
+/*! @brief Test value for the @c additionalHeaders property.
+ */
+static NSString *const kTestAdditionalHeaderValue2 = @"3";
+
 @implementation OIDTokenRequestTests
 
 + (OIDTokenRequest *)testInstance {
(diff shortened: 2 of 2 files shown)


Source/AppAuthCore/OIDTokenRequest.m
```diff
@@ -330,7 +330,7 @@ - (NSURLRequest *)URLRequest {
   }
   
   for (id header in _additionalHeaders) {
-    [URLRequest setValue:httpHeaders[header] forHTTPHeaderField:header];
+    [URLRequest setValue:_additionalHeaders[header] forHTTPHeaderField:header];
```
mdmathias, line 333: Good catch! Would you be able to add a unit test verifying that `-[OIDTokenRequest URLRequest]` has the correct additional parameters?

GravityByte: I wished I could add, but Objective C isn't my native language.

GravityByte: @mdmathias 
I've tried to add test. It works, but I'm not sure that everything has been written correctly and in a right place.

UnitTests/OIDTokenRequestTests.m
```diff
@@ -239,6 +239,10 @@ - (void)testSecureCoding {
   XCTAssertNotNil(request.additionalHeaders, @"");
   XCTAssertEqualObjects(request.additionalHeaders[kTestAdditionalHeaderKey],
                         kTestAdditionalHeaderValue, @"");
+  
+  NSURLRequest *urlRequest = [request URLRequest];
+  XCTAssertEqualObjects([urlRequest.allHTTPHeaderFields objectForKey:kTestAdditionalHeaderKey],
```
mdmathias, line 244: You can remove the third parameter here (`@""`) since it does nothing. I know that is the existing style in this file, but let's start moving away from this practice starting now!

UnitTests/OIDTokenRequestTests.m
```diff
@@ -263,6 +267,10 @@ - (void)testSecureCoding {
   XCTAssertNotNil(requestCopy.additionalHeaders, @"");
   XCTAssertEqualObjects(requestCopy.additionalHeaders[kTestAdditionalHeaderKey],
                         kTestAdditionalHeaderValue, @"");
+  
+  NSURLRequest *urlrequestCopy = [requestCopy URLRequest];
+  XCTAssertEqualObjects([urlrequestCopy.allHTTPHeaderFields objectForKey:kTestAdditionalHeaderKey],
+                        kTestAdditionalHeaderValue, @"");
```
mdmathias, line 273: Same here - no need for the third parameter (`@""`).

UnitTests/OIDTokenRequestTests.m
```diff
@@ -239,6 +239,10 @@ - (void)testSecureCoding {
   XCTAssertNotNil(request.additionalHeaders, @"");
   XCTAssertEqualObjects(request.additionalHeaders[kTestAdditionalHeaderKey],
                         kTestAdditionalHeaderValue, @"");
+
```
mdmathias, line 242: These tests are helpful to add, but they only test this class's [`NSSecureCoding`](https://developer.apple.com/documentation/foundation/nssecurecoding?language=objc) conformance.

I was thinking of also adding a new test case that is specifically focused on ensuring that the additional headers are  present in the token request.

```
- (void)testThatAdditionalHeadersAreInTokenRequest {
  // Construct an `OIDTokenRequest` with `additionalHeaders` 
  // Assert that the `NSURLRequest`'s headers contain what you expect
}
```

This sort of test can verify that the typo is fixed, and can also help to prevent similar typos in the future.

mdmathias (changes requested):

UnitTests/OIDTokenRequestTests.m
GravityByte: @mdmathias 
Please check last commit.

mdmathias (approved):

UnitTests/OIDTokenRequestTests.m
Brooksie18: Thank you. Is this issue resolved or is there something I need to do on my end?

UnitTests/OIDTokenRequestTests.m
Brooksie18: Hello, may I ask if there is anything I can do on my end to help with this issue?

PR #809, opened 2023-12-19
SPM fix depricated warnings

nickolai-nikishin:


mdmathias:
What is the rationale and consequence of this pull request? I don't think we want to suppress warnings, but instead want to fix them.

nickolai-nikishin:
Some projects who is using AppAuth as SPM dependency may have "treat warnings as errors' enabled so the project will fail to compile. Also it brings some 'noise' in build logs. As soon as minimum iOS version will be revisited and deprications will be fixed, this PR can be obsolete

PR #821, opened 2024-02-27
add visionOS support

frogcjn:
This pull request add visionOS support for AppAuth-iOS.
most of the change is happening in adding OIDAuthState and transferred classes from iOS and macOS in the folder `Source/AppAuth/visionOS`.

Package.swift do not declare .visionOS(.v1) since to support compatibility to more lower iOS, macOS versions.


niklas-tag24:
is there any progress here?

PR #825, opened 2024-03-09
Add back missing method to OIDAuthorizationResponse

mdmathias:
AppAuth release 1.7.0 inadvertently removed a method from the public interface of `OIDAuthorizationResponse`. This PR adds the method back.

This fix and subsequent release should fix https://github.com/google/GoogleSignIn-iOS/issues/376.


yalkhder (approved):

RebinX1:
okay so should i just update my packages to the latest version and it should work fine, right?

sengoontoh:
Will there be a release soon?

mdmathias:
@sengoontoh https://github.com/openid/AppAuth-iOS/releases/tag/1.7.1 should address the issue.

Issue #865, opened 2024-06-19
Multiple warnings when building project.


GamaTellez:
**Describe the bug**
Multiple "nil" related warnings (see pic) in multiple classes

**To Reproduce**
Just building project

**Expected behavior**
Only warnings.

**Screenshots**
![Screenshot 2024-06-18 at 9 30 45 PM](https://github.com/openid/AppAuth-iOS/assets/10762590/363c18f3-0789-4cbf-824a-025e62e9c978)

**Environment**
- Device: MacBook Pro Intel
- OS: macOS Sonoma
- Browser: xCode

**Additional context**

First time trying to contribute so let me know anything I can improve in the post.

brnnmrls:
Hi! Thanks for bringing this to our attention. Looks like we were able to see, not the exact issues, but similar issues when we ran "Product > Analyze" on our end. Looks like these are warning indicators that haven't raised any urgent concerns with the library. If you wanted to send a PR fixing some of these issues we'd be happy to take a look.

PR #888, opened 2024-11-27
Fix warnings

loic-sharma:
Address some of the warnings in https://github.com/openid/AppAuth-iOS/issues/865

cc @brnnmrls


Sources/AppAuth/macOS/OIDRedirectHTTPHandler.h
```diff
@@ -60,7 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
         Calling this more than once will result in the previous listener being cancelled (equivalent
         of @c cancelHTTPListener being called).
  */
-- (NSURL *)startHTTPListener:(NSError **)returnError withPort:(uint16_t)port;
+- (nullable NSURL *)startHTTPListener:(NSError **)returnError withPort:(uint16_t)port;
```
loic-sharma, line 63: Comment above indicates this is intentional:

```
@return The URL containing the address of the server with the specified port, or nil if there was an error.
```

PR #917, opened 2025-04-02
Crash fix - when cancelling an authentication session

Prash88:
**Steps to repro**

In Example app,

Tap sign in, hit continue.
Move the app to the background.
Go to Settings > Safari > Clear History and Website Data.
Go into the app again. A new permission dialog is shown.
Tap Cancel. App crashes.

**Recording from example app:** https://streamable.com/5m16cx

**Reasons for crash**

The crash happens because of ASWebAuthenticationSession getting deallocated during or immediately after completion handler
http://ww.openradar.appspot.com/FB12132525

**Fix explanation**

As suggested, this crash can be mitigated by intentionally retaining the authentication session until after the dismiss animation is complete.

### DIFF
--- a/AppAuth.podspec
+++ b/AppAuth.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name         = "AppAuth"
-  s.version      = "2.0.0"
+  s.version      = "1.7.6"
   s.summary      = "AppAuth for iOS and macOS is a client SDK for communicating with OAuth 2.0 and OpenID Connect providers."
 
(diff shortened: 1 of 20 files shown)


brnnmrls:
@Prash88 Do you mind fixing the failing tests, please?

Lomaas:
Why isn't this merged? @Prash88

w-goog:
Thank you for the bump, Simen! We'll get this in the next minor release on a fresh branch (if anyone wants to toss that up, that would speed things along; otherwise I'll handle it 💖)

PR #918, opened 2025-04-07
Update minimum iOS version from 10 to 12.

brnnmrls:
Update the minimum supported iOS version from 10 to 12 and remove any deprecations. 

This PR will remove references relating to [SFAuthenticationSession](https://developer.apple.com/documentation/safariservices/sfauthenticationsession) which is marked deprecated as of iOS 12.
Additionally, this PR will fix deprecation errors coming in relating to [archivedData(withRootObject:)](https://developer.apple.com/documentation/foundation/nskeyedarchiver/archiveddata(withrootobject:)) which has also been deprecated as of iOS 12.

### DIFF
--- a/Sources/AppAuthCore/OIDRegistrationRequest.m
+++ b/Sources/AppAuthCore/OIDRegistrationRequest.m
@@ -130,12 +130,15 @@ static NSString *const kAdditionalParametersKey = @"additionalParameters";
                          forKey:kConfigurationKey];
   NSString *initialAccessToken = [aDecoder decodeObjectOfClass:[NSString class]
                                                         forKey:kInitialAccessToken];
-  NSArray<NSURL *> *redirectURIs = [aDecoder decodeObjectOfClass:[NSArray<NSURL *> class]
-                                                          forKey:kRedirectURIsKey];
-  NSArray<NSString *> *responseTypes = [aDecoder decodeObjectOfClass:[NSArray<NSString *> class]
-                                                              forKey:kResponseTypesKey];
-  NSArray<NSString *> *grantTypes = [aDecoder decodeObjectOfClass:[NSArray<NSString *> class]
-                                                           forKey:kGrantTypesKey];
+  NSArray<NSURL *> *redirectURIs =
+      [aDecoder decodeObjectOfClasses:[NSSet setWithArray:@[[NSArray class], [NSURL class]]]
+                               forKey:kRedirectURIsKey];
+  NSArray<NSString *> *responseTypes =
+      [aDecoder decodeObjectOfClasses:[NSSet setWithArray:@[[NSArray class], [NSString class]]]
+                               forKey:kResponseTypesKey];
+  NSArray<NSString *> *grantTypes =
+      [aDecoder decodeObjectOfClasses:[NSSet setWithArray:@[[NSArray class], [NSString class]]]
+                               forKey:kGrantTypesKey];
   NSString *subjectType = [aDecoder decodeObjectOfClass:[NSString class]
                                                  forKey:kSubjectTypeKey];
   NSString *tokenEndpointAuthenticationMethod =
--- a/UnitTests/OIDAuthStateTests.m
+++ b/UnitTests/OIDAuthStateTests.m
@@ -204,7 +204,18 @@
   NSError *oauthError =
       [[self class] OAuthTokenInvalidGrantErrorWithUnderlyingError:nonCompliantError];
   [authstate updateWithAuthorizationError:oauthError];
-  XCTAssertNoThrow([NSKeyedArchiver archivedDataWithRootObject:authstate], @"");
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    data = [NSKeyedArchiver archivedDataWithRootObject:authstate
+                                 requiringSecureCoding:YES
+                                                 error:&error];
+    XCTAssertNoThrow(data, @"");
+  } else {
+#if !TARGET_OS_IOS
+    XCTAssertNoThrow([NSKeyedArchiver archivedDataWithRootObject:authstate], @"");
+#endif
+  }
 }
 
 /*! @brief Tests @c OIDAuthState.updateWithAuthorizationResponse:error: with a success response.
@@ -358,8 +369,22 @@
   XCTAssert([OIDAuthState supportsSecureCoding], @"");
 
   OIDAuthState *authState = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:authState];
-  OIDAuthState *authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthState *authStateCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    data = [NSKeyedArchiver archivedDataWithRootObject:authState
+                                 requiringSecureCoding:YES
+                                                 error:&error];
+    authStateCopy = [NSKeyedUnarchiver unarchivedObjectOfClass:[OIDAuthState class]
+                                                      fromData:data
+                                                         error:&error];
+  } else {
+#if !TARGET_OS_IOS
+    data = [NSKeyedArchiver archivedDataWithRootObject:authState];
+    authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+#endif
+  }
 
   XCTAssertEqualObjects(authStateCopy.refreshToken, authState.refreshToken, @"");
   XCTAssertEqualObjects(authStateCopy.scope, authState.scope, @"");
@@ -375,9 +400,26 @@
   // Verify the error object is indeed restored.
   NSError *oauthError = [[self class] OAuthTokenInvalidGrantErrorWithUnderlyingError:nil];
   [authState updateWithTokenResponse:nil error:oauthError];
-  data = [NSKeyedArchiver archivedDataWithRootObject:authState];
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    data = [NSKeyedArchiver archivedDataWithRootObject:authState
+                                 requiringSecureCoding:YES
+                                                 error:&error];
+  } else {
+#if !TARGET_OS_IOS
+    data = [NSKeyedArchiver archivedDataWithRootObject:authState];
+#endif
+  }
   XCTAssertNotNil(authState.authorizationError, @"");
-  authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    authStateCopy = [NSKeyedUnarchiver unarchivedObjectOfClass:[OIDAuthState class]
+                                                      fromData:data
+                                                         error:&error];
+  } else {
+#if !TARGET_OS_IOS
+    authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+#endif
+  }
   XCTAssertEqualObjects(authStateCopy.authorizationError.domain,
                         authState.authorizationError.domain, @"");
   XCTAssertEqual(authStateCopy.authorizationError.code, authState.authorizationError.code, @"");
--- a/UnitTests/OIDAuthorizationRequestTests.m
+++ b/UnitTests/OIDAuthorizationRequestTests.m
@@ -322,8 +322,22 @@ static int const kCodeVerifierRecommendedLength = 43;
   XCTAssertEqualObjects(request.additionalParameters[kTestAdditionalParameterKey],
                         kTestAdditionalParameterValue, @"");
 
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request];
-  OIDAuthorizationRequest *requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthorizationRequest *requestCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+    data = [NSKeyedArchiver archivedDataWithRootObject:request
+                                 requiringSecureCoding:YES
+                                                 error:&error];
+    requestCopy = [NSKeyedUnarchiver unarchivedObjectOfClass:[OIDAuthorizationRequest class]
+                                                     fromData:data
+                                                        error:&error];
+  } else {
+#if !TARGET_OS_IOS
+    data = [NSKeyedArchiver archivedDataWithRootObject:request];
+    requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+#endif
+  }
 
   // Not a full test of the configuration deserialization, but should be sufficient as a smoke test
   // to make sure the configuration IS actually getting serialized and deserialized in the
--- a/UnitTests/OIDAuthorizationResponseTests.m
+++ b/UnitTests/OIDAuthorizationResponseTests.m
@@ -158,8 +158,22 @@ static NSString *const kTestScope = @"Scope";
  */
 - (void)testSecureCoding {
   OIDAuthorizationResponse *response = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:response];
-  OIDAuthorizationResponse *responseCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthorizationResponse *responseCopy;
+  NSError *error;
+  NSData *data;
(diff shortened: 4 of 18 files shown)


UnitTests/OIDAuthStateTests.m
```diff
@@ -204,7 +204,18 @@ - (void)testNonCompliantNSCodingNSErrors {
   NSError *oauthError =
       [[self class] OAuthTokenInvalidGrantErrorWithUnderlyingError:nonCompliantError];
   [authstate updateWithAuthorizationError:oauthError];
-  XCTAssertNoThrow([NSKeyedArchiver archivedDataWithRootObject:authstate], @"");
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 209: Curious to know if this check is necessary if we run our tests under an environment where these are guaranteed. Thoughts? Maybe you're thinking about people running this locally with a different environment than we have here...

UnitTests/OIDAuthStateTests.m
```diff
@@ -358,8 +369,22 @@ - (void)testSecureCoding {
   XCTAssert([OIDAuthState supportsSecureCoding], @"");
 
   OIDAuthState *authState = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:authState];
-  OIDAuthState *authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthState *authStateCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 375: Same question here.

UnitTests/OIDAuthStateTests.m
```diff
@@ -375,9 +400,26 @@ - (void)testSecureCoding {
   // Verify the error object is indeed restored.
   NSError *oauthError = [[self class] OAuthTokenInvalidGrantErrorWithUnderlyingError:nil];
   [authState updateWithTokenResponse:nil error:oauthError];
-  data = [NSKeyedArchiver archivedDataWithRootObject:authState];
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 403: Ditto question.

UnitTests/OIDAuthStateTests.m
```diff
@@ -375,9 +400,26 @@ - (void)testSecureCoding {
...
+                                                 error:&error];
+  } else {
+#if !TARGET_OS_IOS
+    data = [NSKeyedArchiver archivedDataWithRootObject:authState];
+#endif
+  }
   XCTAssertNotNil(authState.authorizationError, @"");
-  authStateCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 414: Ditto.

UnitTests/OIDAuthorizationRequestTests.m
```diff
@@ -322,8 +322,22 @@ - (void)testSecureCoding {
   XCTAssertEqualObjects(request.additionalParameters[kTestAdditionalParameterKey],
                         kTestAdditionalParameterValue, @"");
 
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request];
-  OIDAuthorizationRequest *requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthorizationRequest *requestCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 328: Ditto.

UnitTests/OIDAuthorizationResponseTests.m
```diff
@@ -158,8 +158,22 @@ - (void)testCopying {
  */
 - (void)testSecureCoding {
   OIDAuthorizationResponse *response = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:response];
-  OIDAuthorizationResponse *responseCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDAuthorizationResponse *responseCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 164: Ditto.

UnitTests/OIDEndSessionRequestTests.m
```diff
@@ -97,8 +97,22 @@ - (void)testSecureCoding {
     XCTAssertEqualObjects(request.additionalParameters[kTestAdditionalParameterKey],
                           kTestAdditionalParameterValue);
 
-    NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request];
-    OIDEndSessionRequest *requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+    OIDEndSessionRequest *requestCopy;
+    NSError *error;
+    NSData *data;
+    if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 103: Ditto.

UnitTests/OIDRegistrationRequestTests.m
```diff
@@ -142,8 +142,22 @@ - (void)testSecureCoding {
   XCTAssertEqualObjects(request.additionalParameters[kTestAdditionalParameterKey],
                         kTestAdditionalParameterValue);
 
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request];
-  OIDRegistrationRequest *requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDRegistrationRequest *requestCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 148: Ditto.

UnitTests/OIDRegistrationResponseTests.m
```diff
@@ -119,8 +119,22 @@ - (void)testCopying {
  */
 - (void)testSecureCoding {
   OIDRegistrationResponse *response = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:response];
-  OIDRegistrationResponse *responseCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDRegistrationResponse *responseCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 125: Ditto.

UnitTests/OIDServiceConfigurationTests.m
```diff
@@ -363,8 +363,22 @@ - (void)testFetcherWithBadJSON {
  */
 - (void)testSecureCoding {
   OIDServiceConfiguration *configuration = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:configuration];
-  OIDServiceConfiguration *unarchived = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDServiceConfiguration *unarchived;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 369: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -448,21 +448,25 @@ - (void)testSecureCoding {
       [[OIDServiceDiscovery alloc] initWithDictionary:serviceDiscoveryDictionary
                                                              error:&error];
   NSData *data;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 451: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -448,21 +448,25 @@ - (void)testSecureCoding {
...
                                                  error:&error];
   } else {
+#if !TARGET_OS_IOS
     data = [NSKeyedArchiver archivedDataWithRootObject:discovery];
+#endif
   }
   
   OIDServiceDiscovery *unarchived;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 462: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -478,16 +482,18 @@ - (void)testSecureCodingDecodeOld {
       [[OIDServiceDiscoveryOldEncoding alloc] initWithDictionary:serviceDiscoveryDictionary
                                                            error:&error];
   NSData *data;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 485: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -478,16 +482,18 @@ - (void)testSecureCodingDecodeOld {
...
                                                  error:&error];
   } else {
+#if !TARGET_OS_IOS
     data = [NSKeyedArchiver archivedDataWithRootObject:discovery];
+#endif
   }
   
   OIDServiceDiscovery *unarchived;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 496: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -514,16 +522,18 @@ - (void)testSecureCodingOldDecodeNew {
       [[OIDServiceDiscoveryOldDecoding alloc] initWithDictionary:serviceDiscoveryDictionary
                                                            error:&error];
   NSData *data;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 525: Ditto.

UnitTests/OIDServiceDiscoveryTests.m
```diff
@@ -514,16 +522,18 @@ - (void)testSecureCodingOldDecodeNew {
...
                                                  error:&error];
   } else {
+#if !TARGET_OS_IOS
     data = [NSKeyedArchiver archivedDataWithRootObject:discovery];
+#endif
   }
   
   OIDServiceDiscovery *unarchived;
-  if (@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 536: Ditto.

UnitTests/OIDTokenRequestTests.m
```diff
@@ -278,8 +278,22 @@ - (void)testSecureCoding {
   XCTAssertEqualObjects([urlRequest.allHTTPHeaderFields objectForKey:kTestAdditionalHeaderKey],
                         kTestAdditionalHeaderValue);
 
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request];
-  OIDTokenRequest *requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDTokenRequest *requestCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 284: Ditto.

UnitTests/OIDTokenResponseTests.m
```diff
@@ -175,8 +175,22 @@ - (void)testCopying {
  */
 - (void)testSecureCoding {
   OIDTokenResponse *response = [[self class] testInstance];
-  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:response];
-  OIDTokenResponse *responseCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
+  OIDTokenResponse *responseCopy;
+  NSError *error;
+  NSData *data;
+  if (@available(iOS 12.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *)) {
```
mdmathias, line 181: Ditto.

mdmathias (changes requested):

Sources/AppAuthCore/OIDRegistrationRequest.m
```diff
@@ -130,12 +130,15 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
                          forKey:kConfigurationKey];
   NSString *initialAccessToken = [aDecoder decodeObjectOfClass:[NSString class]
                                                         forKey:kInitialAccessToken];
-  NSArray<NSURL *> *redirectURIs = [aDecoder decodeObjectOfClass:[NSArray<NSURL *> class]
-                                                          forKey:kRedirectURIsKey];
-  NSArray<NSString *> *responseTypes = [aDecoder decodeObjectOfClass:[NSArray<NSString *> class]
-                                                              forKey:kResponseTypesKey];
-  NSArray<NSString *> *grantTypes = [aDecoder decodeObjectOfClass:[NSArray<NSString *> class]
-                                                           forKey:kGrantTypesKey];
+  NSArray<NSURL *> *redirectURIs =
```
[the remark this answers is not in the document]
brnnmrls: So this change was added due to some failing tests for `OIDRegistrationRequest` and `OIDRegistrationResponse`. In the case of `OIDRegistrationRequest`, there seemed to be something wrong after unarchiving the object with data since the following tests would fail: https://github.com/openid/AppAuth-iOS/blob/1ba9e0f792210cc6d33ea262f69b7ea53d467323/UnitTests/OIDRegistrationRequestTests.m#L152-L163. These would all be `nil`.

Taking a closer look, it was due to the use of `decodeObjectOfClass` with the bundled classes. Since we're essentially decoding two classes (`NSArray` and `NSURL`) it's necessary to pass in a set of them making it more flexible to handle the decoding. Keeping the previous can be too strict since we expect to match `NSArray<NSString *>` but scenarios could potentially return a `NSMutableArray` instead.

UnitTests/OIDAuthStateTests.m
brnnmrls: Right! I thought about just scratching this check since our environment will guarantee these tests are run, but considering that AppAuth does support older versions than the ones listed here, I think it's worth to have the extra code for the chance we or other devs run this locally with a different environment.

Also considered upping these versions as well but that seems like an additional discussion on top of raising the iOS version.

Sources/AppAuthCore/OIDRegistrationRequest.m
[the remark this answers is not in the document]
mdmathias: Are you saying that these _started_ failing? If so, what was the change that led the old test code to fail? It strikes me that `NSArray<NSURL *>` is both correct and more parsimonious.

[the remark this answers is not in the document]
brnnmrls: Yes, these started failing when updating the deprecated methods to [+archivedData(withRootObject:requiringSecureCoding:) ](https://developer.apple.com/documentation/foundation/nskeyedarchiver/archiveddata(withrootobject:requiringsecurecoding:)) and [+unarchivedObjectOfClass:fromData:error:](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/unarchivedobject(ofclasses:from:)-b9t5) with the introduction of Secure Coding in iOS 12. My guess is that while `NSArray<NSURL *>` correctly defines the property type at compile time, at runtime, the archived object is just an `NSArray`. Secure Coding's strict type checking with `decodeObjectOfClass:` then fails because the runtime class doesn't exactly match `NSArray<NSURL *>`.

`decodeObjectOfClasses:` resolves this by allowing an `NSArray` as the container, while still enabling Secure Coding to verify that its contents are `NSURL` objects. It seems `decodeObjectOfClasses:` is necessary for robust and secure decoding of collections under the stricter runtime checks of Secure Coding.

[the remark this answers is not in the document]
mdmathias: Resolving this because it seems like this is what is required. I am a little surprised that the lightweight generic annotation of `NSArray<NSURL *>` doesn't do what we want, but it seems like that is the case.

mdmathias (approved):

adams6-lap (commented):
Thank you

PR #919, opened 2025-04-18
PR to resolve retain cycle when AppAuth has a strong reference to PresentingViewController

Osama-Fawzi:
We have encountered many cases where there is a retain cycle due to the fact that AppAuth is initialized as a singleton and the login presenting ViewController is referenced strongly by AppAuth.

So, we have switched the reference to be weak so it can be broken when ViewController is dismissed to be deallocated.

also kExpiryTimeTolerance is differentiated between debug and release.


beiglboe:
Would be nice to get this merged

PR #936, opened 2025-08-28
Expose HTTP response headers in `NSError` for non-2xx responses

yilievnet:
### Problem

When a token request returns a non-2xx HTTP status (for example 400), AppAuth currently exposes only the response body and status code in the resulting `NSError`.
There is no way for SDK users to access the response headers, which are required in some flows.

### Solution

Include `HTTPURLResponse.allHeaderFields` in the `NSError.userInfo` for non-2xx responses under a new exported key:

```objc
extern NSString *const OIDHTTPResponseHeadersKey;
```

This change is minimal and fully backward compatible. Successful responses are not affected.

### Background

[DPoP (RFC 9449 §8)](https://datatracker.ietf.org/doc/html/rfc9449#section-8) specifies a case where an authorization server responds to a token request with a 400 status and includes a `DPoP-Nonce` header. Clients must extract this nonce and retry the request with a re-signed proof.

Without access to the response headers, it is impossible to implement a compliant DPoP flow using AppAuth.

### DIFF
--- a/Sources/AppAuthCore/OIDError.h
+++ b/Sources/AppAuthCore/OIDError.h
@@ -73,6 +73,14 @@ extern NSString *const OIDHTTPErrorDomain;
  */
 extern NSString *const OIDOAuthErrorResponseErrorKey;
 
+/*! @brief An error key used to access the HTTP response headers from a failed AppAuth request.
+    @discussion When an AppAuth request fails due to a non-2xx HTTP response, the server's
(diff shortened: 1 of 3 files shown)


guilhermehmcarvalho (approved):

guilhermehmcarvalho:
This is a great addition. As instated by @yilievnet, when you follow the RFC for DPoP, this is a necessary feature. It would be great to see this merged onto the main branch.

Issue #847, opened 2024-04-15
ASWebAuthenticationSession - Current init method is deprecated


adamivancza:
**Is your feature request related to a problem you're having? Please describe.**
Just noticed that with iOS 17.4 Apple depreacted https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/2990952-initwithurl?language=objc

**Describe the solution you'd like**
Use the new init method: https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/4316247-initwithurl?language=objc
At first glance, this would finally allow us to support universal links better as the callback has better support for detecting callback URLs and schemes: https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsessioncallback?language=objc

**Describe alternatives you've considered**

**Additional context**

Issue #367, opened 2019-03-15
Universal link support


galsc-tr:
Hi,

Using SFSafariViewController to login and assuming I've setup Universal Link support for my iOS app on my domain, how do I leverage and redirect to my domain using universal login from within SFSafariViewController?

WilliamDenniss:
You need to call `[_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]` from the method in your delegate that receives the universal link.

E.g. like this, but in the universal link method:

https://github.com/openid/AppAuth-iOS/blob/0c822671e56cedf159a6d27ccf0669b689c0f2b5/Examples/Example-iOS_ObjC/Source/AppDelegate.m#L50-L55

galsc:
How do I redirect from inside the HTML's WebView? (asking for both Android & iOS).

grEvenX:
Is anyone using universal links in production for native apps, I haven't been able to find any examples and I'm struggling a lot with it myself. It works fine the first time (when I don't currently have an auth session and I'm not logged in on the provider) probably because the redirect then happens after a button click and not due to automatic redirection.

See other user with the same problem here:
https://stackoverflow.com/questions/45239215/universal-link-fails-when-oauth2-login-page-redirects-immediately

To describe the issue I'm having:
1. I open the app and it creates an OIDAuthorizationRequest
2. I haven't authorised the app before, so after logging in at the OAuth Provider I see the "consent screen" where I can tap "authorize". The page is then redirected to the redirect URL, and the embedded browser/auth-view is automatically dismissed and I get the access/refresh tokens.
3. The next time I try to create an  OIDAuthorizationRequest I have already authorised the app and I am already logged in at the OAuth Provider, so it automatically redirects be to the redirect URL. When this happens I just see the page for the redirect URL in the embedded browser/auth-view. This is because the OS never forwards the request to the app, thus `application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void`
 is never called.

(Btw, the README and the examples doesn't support Universal Links, because they rely on AppDelegate's `application(_ app: UIApplication, open url: URL, options: [UIApplicationOpenURLOptionsKey : Any] = [:]) -> Bool` instead of the `continue userActivity` as mentioned above).

lapinek:
[...]

PR #938, opened 2025-09-26
support universal link as redirect url

shidoro:
For iOS, when version 17.4+ is available, check whether the redirect url is a universal link. If it is, use the new init method with callback accepting the universal link.

### DIFF
--- a/Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
+++ b/Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
@@ -25,6 +25,7 @@
 #import <SafariServices/SafariServices.h>
 #import <AuthenticationServices/AuthenticationServices.h>
 
+#import <AppAuth/OIDAuthorizationRequest.h>
 #import "OIDErrorUtilities.h"
 #import "OIDExternalUserAgentSession.h"
 #import "OIDExternalUserAgentRequest.h"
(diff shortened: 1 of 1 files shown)


shidoro:
I believe this is related to #847 and #367

weihenglim:
Heads up: If you’re planning to use this, note that HTTPS callbacks require the `webcredentials` service. You’ll need to add it to your AASA and `Runner.entitlements` if you previously only had `applinks`. With that in place, I can confirm this PR works great.

yasmin-ravenwolfe:
I can also confirm - this works great! 

Forked AppAuth-iOS locally with this PR applied, with the `webcredentials` AASA and entitlement config for the redirect url.  The OAuth flow completes end-to-end with session-binding routing the callback back to the originating app. Echoing @weihenglim — works as expected with the webcredentials prerequisite documented in their comment.

Is there a maintainer who could pick this up for review? @w-goog — would love your eyes on this when you have a moment. Happy to help shepherd it through / make any requested changes if that helps unblock.

KhaleelSH:
Any updates on this PR? It would be great if we can have it ASAP. @w-goog

w-goog:
Thank you for the bump, Khaleel! This PR is going to need some work before we merge it, but I'd be happy to try and get it in the post-3.0.0 release. I'll add real comments soon. Quick hits in case someone wants to address them before a formal review:
0. A static class helper - something like `static ASWebAuthenticationSessionCallback *_Nullable
OIDHTTPSCallbackForRequest(id<OIDExternalUserAgentRequest> request)` - would let us write a unit test for this.
1. The unchecked downcast will introduce a crash (try passing `OIDEndSessionRequest` in to `-presentExternalUserAgentRequest`).
2. `prefersEphemeralWebBrowserSession` needs to be set (it's hardcoded to `NO` in this block)
3. Technically this should have a `#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000`, though I am about to bump the version, so, it will be unnecessary by the time we get it in a release.
4. The import should be fixed, it will fail under SPM.
5. We should consider throwing on a malformed redirect, if that's a reachable code path (I wasn't able to test this one yet, but, I think a nil host will hang).
6. `if (@available(iOS 17.4, *) && cond)` is going to be an error with `-Werror` - this needs to be nested. 

I'm sorry for the delay - my workflow previously did not do a good job of catching older PRs / issues that had been bumped. I've addressed that going forward :)

KhaleelSH:
Thank you @w-goog for taking care of this! Are you awaiting changes from the PR author or you'll be handling this?

KhaleelSH:
@w-goog I've created a new PR #978 that supersedes this one and addresses the points you mentioned above.

w-goog:
Thank you for making the new PR! We'll continue the discussion there (probably tomorrow) :)

KhaleelSH:
@w-goog Great! looking forward

PR #952, opened 2026-04-21
Create a new sample app that uses SPM

w-goog:
This PR creates a simple new sample app that uses SwiftUI & SPM. Being unable to do a straightforward build of the various sample apps from the command line motivated me to do so.

I'm guessing there's more to do here - in particular, I'd love to hear if a fresh clone works.


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- Add SwiftUI + Swift Package Manager sample app under `Examples/Example_Swift-SPM`.
```
brnnmrls: Can we link the PR here? Some examples are in the bullets of the previous releases.

Examples/README.md
```diff
@@ -10,6 +10,8 @@ Each example has docs on how to configure:
 
 * [Example for iOS (Objective-C)](Example-iOS_ObjC/README.md)
 * [Example for iOS w/ Carthage (Objective-C)](Example-iOS_ObjC-Carthage/README.md)
+* [Example for iOS w/ Carthage (Swift)](Example-iOS_Swift-Carthage/README.md)
+* [Example for iOS w/ SPM (SwiftUI)](Example_Swift-SPM/README.md)
```
brnnmrls: For consistency, can we rename here and the folder itself to `[Example for iOS w/ SPM (SwiftUI)](Example-iOS_Swift-SPM/README.md)` ?

Examples/Example_Swift-SPM/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+* Issuer
+* Client ID
+* Redirect URI
+
+How to get this information varies by IdP, but we have [instructions](../README.md#openid-certified-providers) for some OpenID Certified providers.
+
+### Configure the Example
+
+This sample reads them from an xcconfig file. Create your local override file by copying the committed defaults:
```
brnnmrls: This may just be a me issue but I got confused when I read 'them' . Let's update to refer to the info above?

Examples/Example-iOS_Swift-SPM/Example/AuthManager.swift
```diff
@@ -0,0 +1,460 @@
...
+                self.logMessage("Error retrieving discovery document: \(error?.localizedDescription ?? "DEFAULT_ERROR")")
+                self.setAuthState(nil)
+                return
+            }
+
+            self.logMessage("Got configuration: \(config)")
+
+            if let clientId = kClientID {
+                self.doAuthWithAutoCodeExchange(configuration: config, clientID: clientId, clientSecret: nil)
+            } else {
```
brnnmrls, line 102: Do we ever enter this else block? Curious because in the init() we call `validateOAuthConfiguration()` which asserts there is a `kClientID`.

Examples/Example_Swift-SPM/Example/AuthManager.swift
```diff
@@ -0,0 +1,460 @@
...
+        super.init()
+        self.validateOAuthConfiguration()
+        self.loadState()
+    }
+
+    // MARK: Public Methods
+
+    func validateOAuthConfiguration() {
+        assert(kClientID != nil, "Register your OIDC Client ID in Example.local.xcconfig (OIDC_CLIENT_ID).")
+        assert(kRedirectURI != "com.example.app:/oauth2redirect/example-provider", "Register your OIDC Redirect URI in Example.local.xcconfig (OIDC_REDIRECT_URI).")
```
brnnmrls: Do we need this check? Asking since there seems to be a similar PreconditionFailure check on line 41.

brnnmrls (changes requested):

Examples/Example_Swift-SPM/Example/AuthManager.swift
w-goog: Removed - Added a reference to the dynamic registration protocol as well, so that it's clear what the branch does.

Examples/Example-iOS_Swift-SPM/Example/AuthManager.swift
w-goog: Correct, the other check was really a bug rather than anything else.

Examples/Example_Swift-SPM/README.md
w-goog: Done :)

brnnmrls (approved):
Looks pretty good! I think it's a pretty good 1:1 mapping from the ObjC example + it makes really good use of the Example.xcconfig and info.plist.

For the future, I'm sure we can add more features if we need it, like explicit token refresh and utilizing the `additionalParameters` argument. Great job!

Issue #953, opened 2026-04-27
[iOS] Support Guided Access Mode


w-goog:
**Is your feature request related to a problem you're having? Please describe.**
AppAuth-iOS is unable to function in Guided Access Mode (unless Safari is allowed, which one might suppose it is generally not). When the ASWAS call fails, the auth flow silently fails.

**Describe the solution you'd like**
Either:
a) The auth flow succeeds in Guided Access Mode
or
b) When the auth flow fails, a clear error is provided, whether to the caller or the user (as appropriate).

**Additional context**
Please see the <tbd> linked PR for additional information.

PR #574, opened 2020-08-14
README changes for AppAuthTV

soule:
Still working on main README changes, included are the changes for the example project/Google IdP details.

PR #954, opened 2026-04-27
[#574] Remove external browser fallback

w-goog:
This PR "fixes" the crash in GSI 9.0 by removing `OIDExternalUserAgentiOS`'s call to `[UIApplication sharedApplication] openURL:requestURL]`. Since the `ASWebAuthenticationSession` can be used for almost all sign-ins, the fallback is much less relevant.

There is one caveat - Guided Access Mode. Before this change, the app would _crash_ when attempting to open the external browser in Guided Access Mode. Now, it fails silently. While this is a strict improvement in behaviour, it offers no real way for AppAuth-iOS to be used by applications utilizing Guided Access Mode. I explored the new API Apple provided, and was unsuccessful in routing to a functioning browser. As this change is still an improvement over the current behaviour, I'll leave additional investigation for future work - filed as issue #953.

Testing: I've tested successful sign-in via the new Swift sample app. I also confirmed that Guided Access Mode replicates the crash; and that this change causes an internal error rather than a crash.

### DIFF
--- a/Sources/AppAuth/iOS/OIDExternalUserAgentIOS.h
+++ b/Sources/AppAuth/iOS/OIDExternalUserAgentIOS.h
@@ -40,9 +40,7 @@ API_UNAVAILABLE(macCatalyst)
 
 /*! @brief The designated initializer.
     @param presentingViewController The view controller from which to present the authentication UI.
-    @discussion The specific authentication UI used depends on the iOS version and accessibility
-        options. iOS 12+ uses @c ASWebAuthenticationSession (unless Guided Access is on),
-        otherwise local browser is used.
+    @discussion The specific authentication UI used depends on the iOS version and accessibility options. Uses @c ASWebAuthenticationSession. If Guided Access is enabled or the session cannot be started, the method returns NO and the authorization flow fails with an error.
  */
 - (nullable instancetype)initWithPresentingViewController:
     (UIViewController *)presentingViewController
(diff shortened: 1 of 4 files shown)


Sources/AppAuth/iOS/OIDExternalUserAgentIOS.h
```diff
@@ -40,9 +40,7 @@ API_UNAVAILABLE(macCatalyst)
 
 /*! @brief The designated initializer.
     @param presentingViewController The view controller from which to present the authentication UI.
-    @discussion The specific authentication UI used depends on the iOS version and accessibility
-        options. iOS 12+ uses @c ASWebAuthenticationSession (unless Guided Access is on),
-        otherwise local browser is used.
+    @discussion The specific authentication UI used depends on the iOS version and accessibility options. Uses @c ASWebAuthenticationSession. If Guided Access is enabled or the session cannot be started, the method returns NO and the authorization flow fails with an error.
```
mdmathias: Please follow the line width we use in this repo. Here and throughout.

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -133,12 +133,9 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
     }
   }
-  // If all else failed use the local browser.
-  if (!openedUserAgent){
-    [[UIApplication sharedApplication] openURL:requestURL
-                                       options:@{}
-                             completionHandler:nil];
-    openedUserAgent = YES;
+  if (!openedUserAgent) {
+    [self cleanUp];
```
mdmathias, line 137: Out of curiosity, would you consider this a breaking change that requires a major version release?

mdmathias (approved):

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.h
w-goog: Updated :)

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
w-goog: I think this specific change would be a patch - "Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior."

I would call this a backwards-compatible bug fix, as the public behaviour is "oopsie, crash".

w-goog: (I'm going to resolve this comment to merge, but please feel free to re-open if you'd like to continue the discussion; or we can do so when we cut a release)

PR #961, opened 2026-06-17
Update runners to macOS 15

w-goog:
The macOS 14 image begins deprecation on July 6th. https://github.com/actions/runner-images/issues/13518

Separately, GitHub removed the simulator runtimes for Xcode <16.3 from macos-15, so the current Xcode 16.2 / iOS 18.2 / tvOS 18.2 pins do not exist on macOS 15, and must thus be updated.

The sister repo GoogleSignIn-iOS runs this same setup on (macos-15 + Xcode 16.4 + iOS 18.6) on its daily builds, so I've updated them to match.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,8 @@ jobs:
     runs-on: macos-15
     strategy:
       matrix:
-        # Check the Github runner images when updating this matrix: https://github.com/actions/runner-images/tree/main/images/macos
+	fail-fast: false
+       	# Check the Github runner images when updating this matrix: https://github.com/actions/runner-images/tree/main/images/macos
         flags: [
           "-scheme AppAuth-iOS -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.6' -sdk 'iphonesimulator'",
           "-scheme AppAuth-macOS -destination 'platform=macOS,arch=x86_64' -sdk 'macosx'",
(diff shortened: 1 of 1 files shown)


brnnmrls (changes requested):
I think the runners are failing on the tvOS jobs: 
`The requested device could not be found because no available devices matched the request.`

Maybe there's some mismatches between our setup and the matrix at https://github.com/actions/runner-images/tree/main/images/macos ?

The error suggests dropping 'OS=18.5' but I didn't look into if that's exactly right.

w-goog:
> I think the runners are failing on the tvOS jobs: `The requested device could not be found because no available devices matched the request.`
> 
> Maybe there's some mismatches between our setup and the matrix at https://github.com/actions/runner-images/tree/main/images/macos ?
> 
> The error suggests dropping 'OS=18.5' but I didn't look into if that's exactly right.

Indeed - I think we might be seeing https://github.com/actions/runner-images/issues/12948, so I've added the download step, and we'll see how that goes :)

w-goog:
> > I think the runners are failing on the tvOS jobs: `The requested device could not be found because no available devices matched the request.`
> > Maybe there's some mismatches between our setup and the matrix at https://github.com/actions/runner-images/tree/main/images/macos ?
> > The error suggests dropping 'OS=18.5' but I didn't look into if that's exactly right.
> 
> Indeed - I think we might be seeing [actions/runner-images#12948](https://github.com/actions/runner-images/issues/12948), so I've added the download step, and we'll see how that goes :)

Ah yep and now I'm indeed seeing the behavior in that issue - "Unable to connect to simulator."

brnnmrls (approved):

PR #966, opened 2026-07-21
Add Swift-specific argument label for resumeExternalUserAgentFlow

### LINKED ISSUE #964: Swift importer generates unusable signature for resumeExternalUserAgentFlow(with:error:)
**Describe the bug**
After the addition of `resumeExternalUserAgentFlowWithURL:error:` in PR #955, the new API appears to be imported incorrectly into Swift.

The Objective-C declaration is:

```
- (BOOL)resumeExternalUserAgentFlowWithURL:(NSURL *)URL
                                     error:(NSError *_Nullable *_Nullable)error;
```

However, Swift resolves the method as if the second parameter has type (), resulting in autocomplete like:

`flow.resumeExternalUserAgentFlow?(with: url, error: ())`

Passing nil or an NSError pointer is not possible, producing errors such as:

`'nil' is not compatible with expected argument type '()'`

As a result, the new API is effectively unusable from Swift, while the deprecated overload still works.

**Screenshots**

Image

**Environment**
- AppAuth 2.1.0
- Xcode 26.6
- iOS 26.5

Suggested fix
Any of:

Make resumeExternalUserAgentFlowWithURL:error: @required (major bump if needed), or
Keep @optional for ObjC but add a Swift overlay / NS_SWIFT_NAME so the documented throwing API is real, or
Update README + Swift examples to stop showing try …(with:) until Swift can call the new method, and document the ObjC-only / wrapper path
Happy to help test a patch.

w-goog:
Fixes #964. Because there's a naming collision between the old & new methods, the NSError**-to-throws special-casing doesn't complete, leaving the real type lost in the `Void`. Which is to say, we were (unfortunately) successful in preserving backwards-compatibility, as the sample app only emitted a warning that the old method was being called. This PR also updates the call in the sample app to ensure this new signature is compiled correctly.

### DIFF
--- a/Examples/Example-iOS_Swift-SPM/Example/AppDelegate.swift
+++ b/Examples/Example-iOS_Swift-SPM/Example/AppDelegate.swift
@@ -29,7 +29,7 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         // such as no pending flow, which previously surfaced as an NSException.
         if let authorizationFlow = self.currentAuthorizationFlow {
             do {
-                try authorizationFlow.resumeExternalUserAgentFlow(with: url)
+                try authorizationFlow.resumeExternalUserAgentFlow?(url)
                 self.currentAuthorizationFlow = nil
                 return true
             } catch {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- Add Swift name to `resumeExternalUserAgentFlowWithURL:error:` and update hint. (#966)
+
 # 2.1.0
 - Add SwiftUI + Swift Package Manager sample app under `Examples/Example-iOS_Swift-SPM`. ([#952](https://github.com/openid/AppAuth-iOS/pull/952))
 - Removed external browser (Safari) fallback from `OIDExternalUserAgentIOS`. If `ASWebAuthenticationSession` fails to start (e.g., Guided Access is enabled), the authorization flow now fails with an error instead of opening an external browser. ([#954](https://github.com/openid/AppAuth-iOS/pull/954))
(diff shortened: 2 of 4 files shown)


Examples/Example-iOS_Swift-SPM/Example/AppDelegate.swift
```diff
@@ -29,7 +29,7 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         // such as no pending flow, which previously surfaced as an NSException.
         if let authorizationFlow = self.currentAuthorizationFlow {
             do {
-                try authorizationFlow.resumeExternalUserAgentFlow(with: url)
+                try authorizationFlow.resumeExternalUserAgentFlow?(url)
```
brnnmrls, line 32: Should this also be updated in the Carthage swift sample: https://github.com/openid/AppAuth-iOS/blob/c4b6c5303f410cca40a200f91cb4649c113105a3/Examples/Example-iOS_Swift-Carthage/Source/AppDelegate.swift#L40 ?

brnnmrls (approved):
LGTM but we should probably open an issue to switch this method to 'required' after allowing clients to migrate. Can we open a new issue or internal tracking bug?

Examples/Example-iOS_Swift-SPM/Example/AppDelegate.swift
w-goog: Done!

w-goog:
Added issue #967 to track the future change.

MiuraKairi:
@w-goog Thanks for the fix! Any ETA on the release?

w-goog:
@MiuraKairi - officially, "soon"! (unofficially, ~EOW?)

w-goog:
Just wanted to update you, @MiuraKairi - ran into a snag with the fix in this PR, so we've decided to do a major version bump and testing is taking a little longer.

PR #968, opened 2026-07-27
Added some protection against adding null objects

Timothybas:
There are cases where the value at this point is null and it crashes the dependent project.


w-goog:
Hi Timothy! Please let us know if there is a bug here we may want to address :)

PR #972, opened 2026-08-12
Bump minimum versions for Xcode 27 support

w-goog:
This PR updates all target minimums + SPM tools version to the minimums required by Xcode 27. It does _not_ do any source cleanup. I've tested with local builds, as Xcode 27 on Github Actions is still in preview.

Xcode 27 will be releasing soon, and we're already doing a major version bump for the new required Swift interface - so, might as well do this version bump too.

Several tvOS tests needed updating, in two groups:
- ~10 tests with an outdated guard, now removed.
- 3 tests using an old API that did not throw. The new API throws, so an error is passed in. No other changes are made to the test, though it may be prudent in a follow-up to assert the error is nil for more helpful test failure messages.


UnitTests/OIDAuthorizationRequestTests.m
```diff
@@ -325,19 +325,12 @@ - (void)testSecureCoding {
...
-#if !TARGET_OS_IOS
-    data = [NSKeyedArchiver archivedDataWithRootObject:request];
-    requestCopy = [NSKeyedUnarchiver unarchiveObjectWithData:data];
-#endif
-  }
+  data = [NSKeyedArchiver archivedDataWithRootObject:request
+                               requiringSecureCoding:YES
+                                               error:&error];
+  requestCopy = [NSKeyedUnarchiver unarchivedObjectOfClass:[OIDAuthorizationRequest class]
+                                                   fromData:data
```
brnnmrls: nit: align colons please

UnitTests/OIDAuthStateTests.m
```diff
@@ -206,16 +206,10 @@ - (void)testNonCompliantNSCodingNSErrors {
   [authstate updateWithAuthorizationError:oauthError];
   NSError *error;
   NSData *data;
```
brnnmrls: can this now just be:
```
NSData *data = [NSKeyedArchiver archivedDataWithRootObject:authstate
                               requiringSecureCoding:YES
                                               error:&error];
```

(with correct alignment lol)

UnitTests/OIDEndSessionRequestTests.m
```diff
@@ -100,19 +100,12 @@ - (void)testSecureCoding {
     OIDEndSessionRequest *requestCopy;
     NSError *error;
     NSData *data;
```
brnnmrls: similar comment as above, can we just set these values when we declare them please?

this comment applies to similar situations below in the other tests

UnitTests/OIDAuthStateTests.m
```diff
@@ -206,16 +206,10 @@ - (void)testNonCompliantNSCodingNSErrors {
...
-    XCTAssertNoThrow(data, @"");
-  } else {
-#if !TARGET_OS_IOS
-    XCTAssertNoThrow([NSKeyedArchiver archivedDataWithRootObject:authstate], @"");
-#endif
-  }
+  data = [NSKeyedArchiver archivedDataWithRootObject:authstate
+                               requiringSecureCoding:YES
+                                               error:&error];
+  XCTAssertNoThrow(data, @"");
```
brnnmrls: This isn't your code, and we can probably investigate this in a different PR, but I'm wondering if the "new" method would ever even throw and whether we should be check for the `error` param as well in these asserts.

brnnmrls (approved):
Approved as long as the above nits are resolved, thanks! Great job!

UnitTests/OIDEndSessionRequestTests.m
w-goog: Done :) (except for the errors, which aren't assigned)

UnitTests/OIDAuthStateTests.m
w-goog: Done :)

UnitTests/OIDAuthorizationRequestTests.m
w-goog: Done :)

UnitTests/OIDAuthStateTests.m
w-goog: Nope, you're correct, great catch! It looks like #918 updated these and missed the functional change in this one - I've added it :)

PR #975, opened 2026-08-20
Fix SPM Sample's install failure

w-goog:
The SPM sample was missing a default bundle ID, which causes the sample to fail to install during testing (when using the default values).


w-goog:
(Swapped reviewer requests so it's easier to do the #974 review)

mdmathias (approved):

Issue #956, opened 2026-05-26
Unable to do logout in iOS


versag1984:
Root Cause Analysis

During logout flow validation, SSO providers (including cloudsso-uat) redirect to the post_logout_redirect_uri without returning the original state parameter.
As AppAuth expects the state value for validation, it reports the error -61439 (State mismatch … but got nil).

More more info:

The issue is a state parameter mismatch during the logout flow between the ON!Track iOS app (using AppAuth-iOS 2.0.0) and the CloudSSO logout endpoint.

The Flow
1. AppAuth-iOS creates an OIDEndSessionRequest with a randomly generated state value and opens the browser to the \
2. CloudSSO LogoutService receives this request, validates the client, and constructs a redirect to Auth0:
https:///v2/logout?returnTo=&post_logo…
The state is embedded as a query parameter inside the returnTo URL (via getLogoutQueryString()).
3. Auth0's /v2/logout terminates the session and redirects the browser to the decoded returnTo URL:
4. AppAuth-iOS intercepts this redirect at https://baseurl/oauth/oauth_callback and parses the URL query parameters looking for state.

The Problem
Looking at the AppAuth-iOS source code (OIDAuthorizationService.m, the OIDEndSessionSession class), the state validation logic is:
if (!OIDIsEqualIncludingNil(_request.state, response.state)) {
// "State mismatch, expecting %@ but got %@ in authorization response %@"
}
The OIDEndSessionResponse is initialized from the query parameters of the redirect URL. The issue is one of these scenarios:

w-goog:
Will you please include the rest of the text? :) It looks like your bug report was cut off.

RazvanMaris:
@w-goog, I'm seeing something similar on my end when trying to logout (where PingFederate is used as an identity server):

State mismatch, 
expecting kpDGD3cKo4ULQAtfJv3b1wWw
but got (null) in authorization response
<OIDEndSessionResponse: 0x11a3c2aco, 
	state:"(null)", 
	additionalParameters: {
	}, 
	request: <OIDEndSessionRequest: 0x1156fdefo, request: [pingFederateURL]/idp/startSLO.ping?id_token_hint=<....>&post_logout_redirect_uri=authgear-oidc-tester://callback&state=kpDGD3cKo4ULQAtfJv3b1wWw
	>
> 

The PingFederate .well-known/openid-configuration file uses /idp/startSLO.ping as the "end_session_endpoint".

Any suggestion on how to address this?

[...]

PR #977, opened 2026-08-25
Update diagnostic errors for missing `state` in logout responses

w-goog:
In issue #956, two different clients (likely) ran into an issue where they received a `nil` state response from the provider.

RFC 6749 section 4.1.2 and OIDC Core 3.1.2.5/6 require the server to echo the state value. Some providers don't handle multi-hop auth (passing the `state` param along) or are doing SAML / proprietary responses (no `state` param). These updates point folks in the right direction.

### DIFF
--- a/Sources/AppAuthCore/OIDAuthorizationService.m
+++ b/Sources/AppAuthCore/OIDAuthorizationService.m
@@ -173,11 +173,11 @@ NS_ASSUME_NONNULL_BEGIN
       if (response.state == nil) {
         userInfo[NSLocalizedDescriptionKey] =
           [NSString stringWithFormat:@"The authorization response is missing the state parameter, "
-                                     "expecting %@. RFC 6749 section 4.1.2 requires the "
-                                     "authorization server to return the exact state value sent "
-                                     "in the request. Authorization response: %@",
-                                     _request.state,
-                                     response];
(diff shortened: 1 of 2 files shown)


Sources/AppAuthCore/OIDAuthorizationService.m
```diff
@@ -307,12 +317,25 @@ - (BOOL)resumeExternalUserAgentFlowWithURL:(NSURL *)URL error:(NSError *_Nullabl
   // verifies that the state in the response matches the state in the request, or both are nil
   if (!OIDIsEqualIncludingNil(_request.state, response.state)) {
     NSMutableDictionary *userInfo = [query.dictionaryValue mutableCopy];
-    userInfo[NSLocalizedDescriptionKey] =
-    [NSString stringWithFormat:@"State mismatch, expecting %@ but got %@ in authorization "
-     "response %@",
-     _request.state,
-     response.state,
-     response];
+    if (!response.state) {
```
brnnmrls, line 320: nit: consider updating to `response.state == nil` to be consisten with line 173 in Sources/AppAuthCore/OIDAuthorizationService.m ?

brnnmrls (approved):

Sources/AppAuthCore/OIDAuthorizationService.m
w-goog: Updated :)

PR #978, opened 2026-09-14
Support HTTPS links as redirect URLs on iOS 17.4+

KhaleelSH:
Supersedes #938, rebased onto 3.0.0 and updated for @w-goog's review there.

For iOS 17.4+, when the redirect URL is an HTTPS universal link, use `ASWebAuthenticationSession`'s HTTPS callback instead of a custom scheme callback.


AppAuth.xcodeproj/project.pbxproj
```diff
@@ -558,6 +558,8 @@
 		A5EEF29A20D821960044F470 /* OIDTokenUtilitiesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A5EEF1FD20CF07760044F470 /* OIDTokenUtilitiesTests.m */; };
 		A5EEF29B20D821970044F470 /* OIDTokenUtilitiesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A5EEF1FD20CF07760044F470 /* OIDTokenUtilitiesTests.m */; };
 		A5EEF29C20D821970044F470 /* OIDTokenUtilitiesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A5EEF1FD20CF07760044F470 /* OIDTokenUtilitiesTests.m */; };
+		AA00AF0100000000000AF001 /* OIDExternalUserAgentIOSTests.m in Sources */ = {isa = PBXBuildFile; fileRef = AA00AF0000000000000AF000 /* OIDExternalUserAgentIOSTests.m */; };
```
w-goog: It looks like these were added by-hand, perhaps when resolving a merge conflict - please delete & re-add via Xcode :)

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -96,11 +119,70 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
+        [session failExternalUserAgentFlowWithError:error];
+        return NO;
+      }
+      __weak OIDExternalUserAgentIOS *weakSelf = self;
+      ASWebAuthenticationSession *authenticationVC =
+          [[ASWebAuthenticationSession alloc] initWithURL:requestURL
+                                                 callback:callback
+                                        completionHandler:^(NSURL * _Nullable callbackURL,
+                                                            NSError * _Nullable error) {
+        __strong OIDExternalUserAgentIOS *strongSelf = weakSelf;
```
w-goog: It looks like this whole block is copied - let's use a shared block, and have the calls only differ on the initializer.

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -96,11 +119,70 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
+      }];
+      authenticationVC.presentationContextProvider = self;
+      authenticationVC.prefersEphemeralWebBrowserSession = _prefersEphemeralSession;
+      _webAuthenticationVC = authenticationVC;
+      openedUserAgent = [authenticationVC start];
+    }
+  }
 
   // iOS 12 and later, use ASWebAuthenticationSession
   if (@available(iOS 12.0, *)) {
```
w-goog: No action requested; just note that you can now remove this @available check to make the callback condensing easier :)

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -41,6 +43,27 @@ @interface OIDExternalUserAgentIOS ()<SFSafariViewControllerDelegate>
 @end
 #endif
 
+API_AVAILABLE(ios(17.4))
+ASWebAuthenticationSessionCallback *_Nullable
```
w-goog: Make this a class method, please :) Then in the test, you can use a category to expose it. See `UnitTests/OIDAuthStateTests.m` for an example.

Sources/AppAuth/iOS/OIDExternalUserAgentIOS.m
```diff
@@ -96,11 +119,70 @@ - (BOOL)presentExternalUserAgentRequest:(id<OIDExternalUserAgentRequest>)request
...
+      if (!callback) {
+        // The redirect URL is HTTPS but a callback couldn't be created for it (e.g. it has no
+        // host). A session started with such a redirect could never complete, so fail instead.
+        [self cleanUp];
+        NSError *error =
+            [OIDErrorUtilities errorWithCode:OIDErrorCodeSafariOpenError
+                             underlyingError:nil
+                                 description:@"The request's HTTPS redirect URL is not a valid "
+                                              "universal link."];
+        [session failExternalUserAgentFlowWithError:error];
```
w-goog: While this is nice, the other failures are left to the caller - let's do the same here. (didFinishWithResponse would clear out the callback of the first call)

UnitTests/OIDExternalUserAgentIOSTests.m
```diff
@@ -0,0 +1,199 @@
+/*! @file OIDExternalUserAgentIOSTests.m
+    @brief AppAuth iOS SDK
+    @copyright
+        Copyright 2025 Google Inc. All Rights Reserved.
```
w-goog: nit: 2026 :)

w-goog (changes requested):
The README says "Both Custom URI Schemes and Universal Links can be used with the library.". That hasn't been true for a while - thank you for fixing it! 

Also, please add "iOS 17.4+, requires the Associated Domains entitlement and an AASA file for the host" to that line of the README :)

KhaleelSH:
Thanks for the review! All addressed.

I added `webcredentials` to your wording, since the HTTPS callback needs the host under that service and apps with only `applinks:` won't work.

Also one behavior change to flag: the shared handler checks the result of `resumeExternalUserAgentFlowWithURL:error:` and fails the flow when it returns NO. `ASWebAuthenticationSession` matches callbacks more loosely than the flow does (a custom scheme matches on the scheme alone, and HTTPS ignores case and the port), so a mismatched redirect used to leave the flow hanging. It now fails with `OIDErrorCodeURLMismatch`, on the custom scheme path too. Happy to limit that to HTTPS if you'd prefer.

w-goog (approved):
Looks good! I updated the language around universal links, since they are technically just HTTPS URLs.

